PR #10, opened 2015-08-13
Add rule for bosh-lite network

georgethebeatle:
Allow 10.0.0.0 network so that apps can see each other in bosh-lite env


jsdelfino:
Thanks! We were doing this in our environments but had forgotten to mention it in the README.
Good to have it documented now.

PR #15, opened 2015-08-21
Module simplifying retrieval of CF token

piotrprzybylski:
Add to the project utility module simplifying retrieval of the Cloud Foundry OAuth token.


jsdelfino:
Thanks for this utility module! Will be useful in all the places where we need to call the CF APIs.

PR #17, opened 2015-08-25
Validate resource_id when submitting usage

georgethebeatle:
[#101946086]

### DIFF
--- a/lib/config/schemas/package.json
+++ b/lib/config/schemas/package.json
@@ -29,7 +29,8 @@
     "pub": "publish"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


jsdelfino:
Thanks for the fix. We'll probably need to adjust this when we provide a 'self service' API allowing service/resource providers to onboard their service/resource definitions themselves as that'll open up the resource_id enumType and will require a DB query or call to a 'resource definition' service to actually validate the resource, but what you've done here will work well for now. Thanks again!

jsdelfino:
Merging as looks like it's possible to merge that pull request again now. I believe that Github was having some issues earlier today, but things look OK now.

PR #20, opened 2015-08-29
Token validation and decoding, extracting user info and scopes

piotrprzybylski:
Update to enable token validation and decoding, extracting from the token the user information - id and name and scopes array.


sasrin:
Thank you for useful additional functionalities.

PR #22, opened 2015-09-09
Add CF bridge prototype

hsiliev:


jsdelfino:
Thank you for this nice contribution! Abacus needs to be able to meter CF apps in addition to services and this is an important step towards being able to provide that feature.

PR #23, opened 2015-09-10
Improve container resource reporting

georgethebeatle:
* Report memory in GB/Hour for container resource
* Fix reporting with no time parameter


jsdelfino:
Merging this pull request. Thanks for the improvements!

I also added a few comments to the commits, as we may need a few more minor changes on top of these changes, after I merge. Let me know what you think.

jsdelfino:
@hsiliev I had missed the second route in my initial review. I got it now, that makes sense :) Thanks!

jsdelfino:
@hsiliev, OK I'll adjust the measurement units to MB.

PR #27, opened 2015-09-12
Cover cf-bridge with unit tests

georgethebeatle:
* Provide unit tests for the new bridge components
* Tweak mocha configuration so that it executes every file ending with `test.js` in the `test/` dir

### DIFF
--- a/lib/cf/bridge/src/index.js
+++ b/lib/cf/bridge/src/index.js
@@ -1,7 +1,5 @@
 'use strict';
 
-/* istanbul ignore file */
-
 const router = require('abacus-router');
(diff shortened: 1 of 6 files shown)


georgethebeatle:
Unfortunately I can't think of any more test cases that make sense...

georgethebeatle:
Added one more case for empty event stream coming from CF API. Let's see if it's gonna make it.

jsdelfino:
@hsiliev, @georgethebeatle Looks good. Thanks for the tests and the useful addition to the Mocha test tools.

PR #28, opened 2015-09-14
Handle failed report request to Abacus

georgethebeatle:
* Keep usage events in memory when POST to Abacus fails, so that reporting is retried on next iteration
* Refactor reporting logic
[#101018516]


jsdelfino:
Thanks for this contribution!

PR #29, opened 2015-09-15
Avoid race conditions on usage events read

georgethebeatle:
Events are read with old GUID set far back in time (1 min)
https://www.cloudfoundry.org/how-to-bill-on-cloud-foundry/

[#101018516]


jsdelfino:
Looks good to me. Thanks!

Issue #36, opened 2015-09-22
Unable to push to cloudfoundry.


rajkiranrbala:
I was trying to deploy abacus to a local cloud foundry infrastructure and the deployment failed while preparing to push the rating application.

When I do this
```sh
$ npm run cfpush
```

I get the following error
```sh
...
+ cd lib/aggregation/rate
+ npm run cfpush -- -i 1
npm ERR! Darwin 14.5.0
npm ERR! argv "node" "/Users/rajkiranrbala/workspace/cf-abacus/node_modules/.bin/npm" "run" "cfpush" "--" "-i" "1"
npm ERR! node v0.10.39
npm ERR! npm  v2.10.1

npm ERR! missing script: cfpush
npm ERR! 
npm ERR! If you need help, you may report this error at:
npm ERR!     <https://github.com/npm/npm/issues>

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/rajkiranrbala/workspace/cf-abacus/lib/aggregation/rate/npm-debug.log

npm ERR! cf-abacus@0.0.1-dev.3 cfpush: `bin/cfpush`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the cf-abacus@0.0.1-dev.3 cfpush script.
npm ERR! This is most likely a problem with the cf-abacus package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     bin/cfpush
npm ERR! You can get their info via:
npm ERR!     npm owner ls cf-abacus
npm ERR! There is likely additional logging output above.
npm ERR! System Darwin 14.5.0
npm ERR! command "/usr/local/bin/node" "/usr/local/bin/npm" "run" "cfpush"
npm ERR! cwd /Users/rajkiranrbala/workspace/cf-abacus
npm ERR! node -v v0.10.39
npm ERR! npm -v 1.4.28
npm ERR! code ELIFECYCLE
npm ERR! 
npm ERR! Additional logging details can be found in:
npm ERR!     /Users/rajkiranrbala/workspace/cf-abacus/npm-debug.log
npm ERR! not ok code 0

```

rajkiranrbala:
I did see that the rate application is missing the following
> manifest.yml file
> cfpush script in package.json file

jsdelfino:
Closing as this is now fixed.

jsdelfino:
[...]

PR #42, opened 2015-09-23
Fixed rate application cfpush issue

rajkiranrbala:
Fixed https://github.com/cloudfoundry-incubator/cf-abacus/issues/36 by adding manifest file and updating the package.json file.


jsdelfino:
The changes look good to me, although I'm not too sure about that second merge commit 44ed1b3 (I think it'll be ignored when I merge with master, otherwise you'll have to redo that pull request without merge, i.e. from a clean up to date version of the upstream master in your fork).

jsdelfino:
The merge worked. Thanks for the fix!

PR #83, opened 2015-10-13
APIs to retrieve resource instance level accumulated usage.

rajkiranrbala:
APIs to retrieve resource instance level accumulated usage

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -19,6 +19,8 @@ Usage summary report
 GraphQL usage query
 
 Resource usage
+
+Resource Instance Usage
 ---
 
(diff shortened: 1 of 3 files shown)


jsdelfino:
Thanks for this initial contribution, which looks like a good start.

I'd love to get that feature in Abacus at some point but I'd like to see if it'd be possible to integrate the navigation from aggregated usage to individual resource-instance level usage in a GraphQL navigation instead of a completely different API. 

Also, we may want to discuss this topic on the cf-dev the mailing list as I'm wondering if some changes to the Abacus database schemas would help make the implementation of that type of query a bit easier.

Thanks!

PR #85, opened 2015-10-13
Read URI env from CF service instances

hsiliev:
[#101019050]

CF Service instance URI are used to resolve URIs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .DS_Store
 README.html
+CHANGES.html
 node_modules/**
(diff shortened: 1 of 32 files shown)


jsdelfino:
@hsiliev, @georgethebeatle: Thanks for this addition to resolve the uris from CF service instances! Not sure what the other 14 commits showing in the pull request were (looks like a merge into another branch...) but although they now show in the project's Github history (as duplicates of previous commits I guess...) fortunately I don't think they made it into the real Git history.

PR #172, opened 2015-11-18
fixed a bug with perf - healthy

KRuelY:


jsdelfino:
Good catch. Thanks for the contribution!

Issue #175, opened 2015-11-19
Identify what pricing was used to rate usage


rajkiranrbala:
Is it better to have the accountId to be included as a part of the rated usage doc, to identify what pricing group was applied while the  aggregated usage doc was rated?

jsdelfino:
Identifying what pricing was applied to rate the usage would make a lot of sense to me. I'm not sure that adding the account id to the usage doc achieves that goal though, as the pricing could evolve over time or be shared across multiple accounts for example. Also, an org belongs to an account so adding the account id on top of the org id which is already in the doc doesn't really provide any useful info as I understand it. So, instead I'd suggest to add an id to the price config, and include that id in the rated usage doc.

jsdelfino:
I changed the issue subject to reflect the described requirement

jsdelfino:
The id of the pricing plan used to rate usage is now included in the rated usage docs.

PR #178, opened 2015-11-23
Includes account_id as a part of the rated document

rajkiranrbala:
Included account_id as a part of the rated usage doc. https://github.com/cloudfoundry-incubator/cf-abacus/issues/175

### DIFF
--- a/demo/client/src/test/test.js
+++ b/demo/client/src/test/test.js
@@ -161,6 +161,7 @@ describe('abacus-demo-client', () => {
     // Expected usage report for the test organization
     const report = {
       organization_id: 'us-south:a3d7fe4d-3cb1-4cc3-a831-ffe98e20cf27',
+      account_id: '1234',
       windows: buildWindow(46.09),
       resources: [{
         resource_id: 'object-storage',
@@ -344,4 +345,3 @@ describe('abacus-demo-client', () => {
     });
   });
 });
-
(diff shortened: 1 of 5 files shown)


jsdelfino:
Looking at this pull request. I'll build it locally as the Travis build failed, but not sure if it's related to these changes or not.

@rajkiranrbala Can you help us understand the purpose of storing the account id in the rated usage doc?

@sasrin Should we wait for your refactoring of the aggregator and rating services to merge this in? Not sure yet if the account id should be included in the doc stored in the db, but if we decide that it needs to be included then I'm thinking that the code change to include it will look pretty different once you're done with your refactoring... What do you think?

sasrin:
> @sasrin Should we wait for your refactoring of the aggregator and rating services to merge this in? Not sure yet if the account id should be included in the doc stored in the db, but if we decide that it needs to be included then I'm thinking that the code change to include it will look pretty different once you're done with your refactoring... What do you think?

@jsdelfino  - We would need to do a merge...but it looks like it's going to be a straight forward merge, so we should be OK.

jsdelfino:
See my latest comment in #175. I'm not convinced that adding the account id to the rated usage doc is really useful. I'm suggesting to add a price config id instead.

rajkiranrbala:
There are possibilities where an organization can be moved between accounts. So whenever an aggregated doc is rated at org level, including account_id as a part of the doc helps in identifying the target account towards which the cost is incurred.

PR #186, opened 2015-12-02
Changed the DEBUG so it will not print out db passwords anymore

AzureEnigma:
This change makes it so when credidentials are in the url, it will not print them in the debug mode anymore


jsdelfino:
Going to merge this one and maybe apply another change on top of it as it looks like the protocol won't be printed anymore with that change. Thanks for the contribution!

jsdelfino:
Changed how we obscure the user and password from the DB URI using a regular expression instead of a split() to preserve the rest of the URI. Also covered another DEBUG call in the `dbclient` module which was still printing the user and password. See commit c451fc7 for that change.

PR #190, opened 2015-12-14
Aggregate doc with account_id

rajkiranrbala:
Includes account_id as a part of the aggregated doc. https://github.com/cloudfoundry-incubator/cf-abacus/issues/175


jsdelfino:
Looks OK to me, merging. Thanks for the contribution!

PR #191, opened 2015-12-15
Resolve Org's Account Based on Time

rajkiranrbala:
Since the organizations can be moved across accounts, the look up for an org's account should have a time parameter.


jsdelfino:
Closing this as similar function has now been implemented in a different way as part of the recent refactoring of the account service.

PR #201, opened 2016-01-14
Added time to the api to get account information using organization_id

KRuelY:


jsdelfino:
Looks good to me now. Merging.

PR #203, opened 2016-01-15
Account stub: introduce new API that returns rating config.

KRuelY:
The schema for rating_config is up for discussion.

The collector would call this API to 'validate', get the rating_plan_id, and attach the rating_plan_id to the usage document. 

Currently this API returns the whole rating_configuration, and collector would _.pick the rating_plan_id. 

I think we should have another API that would returns the rating config given a rating_plan_id and time because it seems unnatural for accumulator, aggregator, and reporting to get rating config by passing org_id, resource_id, resource_type, plan_id, and time when we already have the rating_plan_id from the collector. If we are to implements this another API, then the validate API should just returns the rating_plan_id.

### DIFF
--- a/lib/config/schemas/src/index.js
+++ b/lib/config/schemas/src/index.js
@@ -10,6 +10,7 @@ const resourceConfig = require('./resource-config.js');
 const priceConfig = require('./price-config.js');
 const resourceUsage = require('./resource-usage.js');
 const organizationReport = require('./organization-report.js');
+const ratingConfig = require('./rating-config.js');
 
 // Compile a type into a JSON schema, GraphQL schema and validate function
 const compile = (type) => {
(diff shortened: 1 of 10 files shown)


jsdelfino:
I added a few comments to your commit. Let me know what you think.

Also w.r.t
> I think we should have another API that would returns the rating config given a rating_plan_id and time 

+1 to have that other API, although I'm not sure we even need the time, as if the rating config changes over time then it's rating plan id will change as well. Makes sense?

jsdelfino:
For some reason, this pull request seems to conflict with the master branch. That may be easy to fix anyway when you resubmit the changes addressing my comments :) Thanks!

KRuelY:
Hi Sebastien. I made the necessary changes and addressed the conflict with the master branch. Thanks!

jsdelfino:
Thanks!

PR #204, opened 2016-01-15
New API in the account stub: returns pricing_plan_id and prices

KRuelY:
The intent of this API is for the collector to:
- attach prices to the usage document to avoid multiple price lookup
- attach pricing plan id to the usage document to allow accumulation to pricing_plan_id

This is the draft of the API. If this schema is good, I'll go ahead and add pricing schema, caching, and all the necessary adjustment.

### DIFF
--- a/lib/stubs/account/src/index.js
+++ b/lib/stubs/account/src/index.js
@@ -12,6 +12,9 @@ const oauth = require('abacus-oauth');
 const schemas = require('abacus-usage-schemas');
 
 const extend = _.extend;
+const pick = _.pick;
+const map = _.map;
+const filter = _.filter;
 
 /* jshint noyield: true */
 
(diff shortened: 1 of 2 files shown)


jsdelfino:
Had one minor comment but the rest looks good to me so merging. Hoping you can address my comment in a subsequent commit. Thanks!

KRuelY:
That is one thing that I want to discuss, the schema of prices that will be attached to the usage doc.

The mutation and data movement is to format the prices to become:
prices: [{
  name: 'storage',
  price: 1.00
}, {
  name: 'thousand_light_api_calls',
  price: 0.03
}]

instead of attaching the whole doc from: https://github.com/cloudfoundry-incubator/cf-abacus/blob/master/lib/stubs/account/src/resources/object-storage.js

I'm thinking that we should introduce a new pricing config schema with pricing_plan_id, but I didn't do that because it'll break the pipeline. Should I add a new pricing config schema on different folder for now just to get some feedback?

Thanks, I'll get started on the collector once the price schema is decided.

PR #213, opened 2016-01-27
Add mongodb client module

hsiliev:


hsiliev:
Integration tests fail due to the ongoing refactoring. Created #214 to track the problem.

Merging the new module since it is unrelated to the problem.

lib/utils/mongoclient/src/test/keys/server.key
```diff
@@ -0,0 +1,18 @@
+-----BEGIN RSA PRIVATE KEY-----
```
maximilien: @hsiliev why are private keys and certs being committed in the clear?

Even if they are no longer valid, I'd suggest we use some place holder. Can we do that?

hsiliev: @maximilien We use a self-generated CA to self-sign client and server certs and keys. Then we use them to start a test-server and connect with the mongo client. The files are used only as test resources.

PR #226, opened 2016-02-03
Removed duplicate key "babel-preset-es2015" in "dependencies"

danparsons:
The duplicate key causes "npm run lint" to fail.


jsdelfino:
Good catch. Thanks!

PR #227, opened 2016-02-05
Retrieve oauth client ID and secret from env vars

danparsons:
Updated README.md too.


danparsons:
I see where the build failed, I'll fix soon and re-submit

PR #229, opened 2016-02-08
Retrieve oauth client ID and secret from env vars

danparsons:
Updated README.md to reflect this too. This commit includes a fix for the previous PR that failed build tests.

### DIFF
--- a/lib/cf/bridge/src/index.js
+++ b/lib/cf/bridge/src/index.js
@@ -293,7 +293,9 @@ const stopReporting = () => {
 };
 
 const scheduleUsageReporting = () => {
-  oauth.start('bridge', 'secret');
+  var clientId = process.env.CLIENT_ID ? process.env.CLIENT_ID : "bridge";
+  var secret = process.env.SECRET ? process.env.SECRET : "secret";
+  oauth.start(clientId, secret);
 
   // Start purge compensation after we are sure we have stable event order
   module.appsReader = setTimeout(fetchCFApps, 0);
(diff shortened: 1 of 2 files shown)


danparsons:
Whoops! I shall try again...

lib/cf/bridge/src/index.js
```diff
@@ -293,7 +293,9 @@ const stopReporting = () => {
 };
 
 const scheduleUsageReporting = () => {
-  oauth.start('bridge', 'secret');
+  var clientId = process.env.CLIENT_ID ? process.env.CLIENT_ID : "bridge";
```
hsiliev: I personally would prefer to have no hardcoded user/pass. If you go this direction you will need to modify the cf-bridge integration test in tests/cf/bridge

danparsons: I would prefer that too, but it had hard coded values originally, I assumed they were there on purpose?

hsiliev: This was the easiest way to have the bridge running. We had the idea to make this configurable and use the oauth module in abacus (https://www.pivotaltracker.com/story/show/108061970).

lib/cf/bridge/src/index.js
```diff
@@ -293,7 +293,9 @@ const stopReporting = () => {
...
+  var secret = process.env.SECRET ? process.env.SECRET : "secret";
```
hsiliev: You can use `const` instead of `var`

PR #254, opened 2016-02-25
Update webhook module to allow publication

csaroff:
Publishers can create webhooks and supply HMAC authentication to their subscribers.

### DIFF
--- a/lib/utils/webhook/README.md
+++ b/lib/utils/webhook/README.md
@@ -3,3 +3,38 @@ abacus-webhook
 
 Wrapper around the request module providing an easy way to implement a Webhook.
 
+Make an http post request to a list of subscriptions.  This module behaves
+similarly to [request](https://github.com/request/request) except that it
+takes an array of endpoints as input and its default request type is post
+instead of get.
+
+Usage
+---
+
+Post a message to a list of subscribers
+
+```javascript
+  const webhook = require('abacus-webhook');
+
+  const endpoints = ['https://foo.com/bar', 'https://bar.com/foo'];
+
+  const body = {
+    event: 'account_region_move',
+    account_id: '5d32efa8-8572-4388-b026-aebddc7e42c7',
+    old_pricing_country: 'USA',
+    new_pricing_country: 'CAD'
+  };
+
+  const options = {
+    json: body,
+    // Send messages one at a time.
+    concurrency: 1
+  };
+
+  webhook(endpoints, options, (error, responses) => {
+    if(!error)
+      responses.map((response) => {
+        // Do more stuff
+      });
+  });
+```
--- a/lib/utils/webhook/package.json
+++ b/lib/utils/webhook/package.json
@@ -30,15 +30,16 @@
   },
   "dependencies": {
     "abacus-debug": "file:../debug",
-    "abacus-request": "file:../request",
+    "abacus-throttle": "file:../throttle",
     "abacus-transform": "file:../transform",
     "babel-preset-es2015": "^6.6.0",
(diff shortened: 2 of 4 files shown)


jsdelfino:
Hey @csaroff could you resubmit this as a single commit maybe? It's a little difficult for me to review the latest state as changes are distributed in different commits. Thanks!

csaroff:
Squashed.  Do you prefer all PR's submitted as one commit?

jsdelfino:
I'm going to open an issue to discuss Webhook authentication, as it'll be easier for others to follow (and I'm also not able to figure out how to respond to your comment response here...)

csaroff:
This update resolves issue #261 by allowing abacus-webhook to remain entirely agnostic to a module consumer's use of authentication.  It uses [request](https://github.com/request/request) instead of abacus-request because abacus-request lacks the [request.defaults](https://github.com/request/request#requestdefaultsoptions) method(pending a resolution to issue #272).

betafood:
Hey @csaroff, everything looks good from a quick look through it. I do have one question. I do understand and agree with not using abacus-request due to the current lack of the defaults function, but the question I have is based on the options passed here:
```
const request = require('request').defaults({ method: 'post' });
```
I'm wondering. Wouldn't it also work with abacus request if you replace this line:
```
const trequest = throttle(request, options.concurrency || 5);
```
to say
```
const trequest = throttle(request.post, options.concurrency || 5);
```
and not worry about setting a defaults?

csaroff:
@betafood Yes, it will work, but it forces the API consumer to use http post.  While this is the standard http verb used in webhooks, I was hoping to give the api consumer the ability to change the http verb if they wanted to.  I think that the long term solution is to make abacus-request more extensible, but in the meantime, do you have any wisdom here?  Should I just update the module to be more strict and use abacus-request?

betafood:
@csaroff I think I agree with your reasoning here. I only brought it up since I only saw the post passed into the options. The idea is that a consumer would be able to pass in some sort of options whether it's just the http verb to the request defaults as well, correct?

I personally don't see a problem with using request for now until abacus-request has its own defaults function implemented (as you said would be the long term solution here). The only thing that pops out to me at the moment is losing debugging with an variable like "abacus-*" or "e-abacus-*"

csaroff:
> The idea is that a consumer would be able to pass in some sort of options whether it's just the http verb to the request defaults as well, correct?
Correct

> The only thing that pops out to me at the moment is losing debugging with an variable like "abacus-" or "e-abacus-"
That's a good point.  I'm happy to update to abacus-request.  I don't have a strong opinion on extensibility vs. debugging advantages here.  @KRuelY Do you have any insight?

KRuelY:
> That's a good point. I'm happy to update to abacus-request. I don't have a strong opinion on extensibility vs. debugging advantages here.

I do not have a strong opinion too, but if I have to say something:

The abacus-request debug is probably the most useful debug. It is the only debug set (Other than e-*) in the integration test that I have because it tells us the target, statusCode, body, options, etc.

With abacus-request's debug, we do not have to write the debug code block(Something like 'making request to this endpoint with body x' and 'the result of the call is x') each time we want to make a request call in the app/module that we're working on. Which could be convenient for the webhook module in case you want to see the result of the calls.

We probably want both extensibility and the convenience for debugging, so it is probably better to update abacus-request?

csaroff:
@KRuelY Based on what you're saying, I gather that a PR for #272 is a prerequisite for this PR.  Thanks for the explanation.

maximilien:
@KRuelY and @csaroff any updates on this PR?

Issue #272, opened 2016-03-08
abacus-request provides no defaults method.


csaroff:
The popular request library that abacus-request is based on currently provides a method, [defaults](https://github.com/request/request#requestdefaultsoptions), which returns a new request object with the given options set as default for each call to request.  abacus-request should implement this method to allow module consumers to do things like set the default http method to post.

```javascript
const request = require('abacus-request').defaults({ method: 'post' });
```

or to set default http headers.

```javascript
const request = require('abacus-request').defaults({
  headers: {'x-token': 'my-token'}
});
```

jsdelfino:
The Abacus request module also uses request.defaults internally, so I'm not sure about the effects of other calls to request.defaults on top of that (in particular I'm wondering if that'll still allow a shared connection pool to be used by all instances of Abacus request...) If somebody wants to provide an implementation of the defaults function I'll be happy to review it but I'll want to see some tests showing that these defaults can be combined with the internal Abacus request defaults.

jsdelfino:
Closing this issue until somebody contributes an acceptable pull request for it. Please reopen if you feel like contributing such a pull request. Thanks!

PR #280, opened 2016-03-18
Changed the request module to allow customizaiton of defaults

AzureEnigma:
The request module will allow the user to change the defaults that the abacus-request module uses when it uses the request module. Tests still pass, but have not written any new tests for the new change.

Really submitted this pull request to see if this was the correct way to approach this problem.

### DIFF
--- a/lib/utils/request/src/index.js
+++ b/lib/utils/request/src/index.js
@@ -27,6 +27,16 @@ const omit = _.omit;
 const debug = require('abacus-debug')('abacus-request');
 const edebug = require('abacus-debug')('e-abacus-request');
 
+// The default request options
+let defopts = {
+  json: true,
+  rejectUnauthorized: false,
+  forever: true,
+  pool: {
+    maxSockets: 100
+  }
+};
+
 // Return the list of parameters found in a URI template
 const params = (template) => {
   return map(template.match(/:[a-z_][a-z0-9_]*/ig), (k) => k.substr(1));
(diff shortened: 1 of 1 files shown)


lib/utils/request/src/index.js
```diff
@@ -27,6 +27,16 @@ const omit = _.omit;
 const debug = require('abacus-debug')('abacus-request');
 const edebug = require('abacus-debug')('e-abacus-request');
 
+// The default request options
+let defopts = {
```
jsdelfino: defopts shouldn't be mutable, the behavior of request.defaults is to return a new instance of request loaded with the defaults, as opposed to changing the defaults for other usages of request, which could be very disruptive in an environment with many unrelated users of the request module

lib/utils/request/src/index.js
```diff
@@ -42,15 +52,15 @@ const route = (template, parms) => {
...
 // Setup request default options
-const drequest = request.defaults({
-  json: true,
-  rejectUnauthorized: false,
-  forever: true,
-  pool: {
-    maxSockets: 100
-  }
-});
+const drequest = request.defaults(defopts);
```
jsdelfino: This will be executed before the defaults are set by your setDefaults function

lib/utils/request/src/index.js
```diff
@@ -42,15 +52,15 @@ const route = (template, parms) => {
   });
 };
 
+// Changes the request defaults abacus uses
+// Will override the same options or add the new one
+const setDefaults = (opts) => {
+  defopts = extend(defopts, opts);
+  return xrequest;
```
jsdelfino: AIUI the behavior you've coded here would be very different from the original behavior of request.defaults (which returns an instance of request pre-configured with the given defaults)

jsdelfino:
This pull request needs more work, see my comments in the code.

jsdelfino:
Also see my comments in issue #272

AzureEnigma:
Sorry for the noob questions, but do you have any recommendations for the approach of being able to add defaults to request?

- Should I move the request module into xrequest? It seems like each require of the abacus-request module should have its own request to modify.

- What would be the best way of modifying the defaults of the request? Should I have an extra parameter for xrequest to see if there are extra default options?

jsdelfino:
Closing for now as the pull request can't be merged as is. Please reopen or open another one addressing  my comments. Thanks!

PR #284, opened 2016-03-25
Secure /healthcheck and /hystrix.stream with basic authentication.

KRuelY:
Added basicAuth to healthcheck and hystrix.stream.

I would be happy to make changes on the basic authentication. Currently this is what the basic authentication is doing:

It takes in basic auth token and converts it to username and password. It then call UAA to get bearer token along with the scopes that the user has. It then checks whether the user has the scope 'abacus'.

What troubles me:
the UAA plugins that we have requires the caller to pass in the scope. The token that will be returned by the UAA plugins would have the scopes that was passed. This means that I cannot simulate unauthorized user and authorized user. Therefore I made it so that the only user that can pass basic auth would need to have
username: process.env.USERNAME || 'abacus'
password: process.env.PASSWORD || 'HS256'

Any feedback to solve this problem would be appreciated.

Thanks!

### DIFF
--- a/lib/utils/eureka/src/index.js
+++ b/lib/utils/eureka/src/index.js
@@ -5,32 +5,37 @@
 const request = require('abacus-request');
 const perf = require('abacus-perf');
 const urienv = require('abacus-urienv');
+const oauth = require('abacus-oauth');
 
 // Setup debug log
 const debug = require('abacus-debug')('abacus-eureka');
 
+// Return the Eureka v2 API URL. Can be configured with a EUREKA env variable
+// and default to http://localhost:8080.
+const uris = urienv({
+  eureka: 9990,
+  auth_server: 9882
+});
+
 // Return an Express middleware that responds to a health check request
 const health = () => {
   return function(req, res, next) {
-    if(req.path == '/healthcheck') {
-
-      // Get the health of the app computed from the perf stats
-      const h = perf.healthy();
-      debug('Returning app health %s', h);
-      res.status(h ? 200 : 500).send({
-        healthy: h
-      });
-    }
+    if(req.path == '/healthcheck')
+      oauth.basicAuth(uris.auth_server, req.headers &&
+        req.headers.authorization, (err, authenticated) => {
+          if (err)
+            throw err;
+          // Get the health of the app computed from the perf stats
+          const h = perf.healthy();
+          debug('Returning app health %s', h);
+          res.status(h ? 200 : 500).send({
+            healthy: h
+          });
+        });
     else
       next();
   };
 };
-
-// Return the Eureka v2 API URL. Can be configured with a EUREKA env variable
-// and default to http://localhost:8080.
-const uris = urienv({
-  eureka: 9990
-});
 
 const server = (host) => {
   const s = host ? host : process.env.EUREKA ? uris.eureka : undefined;
--- a/lib/utils/eureka/src/index.js
+++ b/lib/utils/eureka/src/index.js
@@ -7,6 +7,8 @@ const perf = require('abacus-perf');
 const urienv = require('abacus-urienv');
 const oauth = require('abacus-oauth');
 
+const secured = () => process.env.SECURED === 'true' ? true : false;
+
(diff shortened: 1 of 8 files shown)


lib/utils/eureka/src/index.js
```diff
@@ -5,33 +5,38 @@
...
-
-      // Get the health of the app computed from the perf stats
-      const h = perf.healthy();
-      debug('Returning app health %s', h);
-      res.status(h ? 200 : 500).send({
-        healthy: h
-      });
-    }
+    if(req.path == '/healthcheck')
+      oauth.basicAuth(uris.auth_server, req.headers &&
```
jsdelfino: This should pass in the scopes that we're expecting the user to have for this particular route.

jsdelfino: The oauth.basicAuth function seems to bundle 3 unrelated capabilities:
- extracting the user and password from the Authorization header
- authenticating the user using UAA
- authorizing the user to perform something (here, perform a health check)

I think this should translate to 3 different functions.

lib/utils/eureka/src/index.js
```diff
@@ -5,33 +5,38 @@
...
+        req.headers.authorization, (err, authenticated) => {
```
jsdelfino: The user is not only authenticated here (in reference to the authenticated parameter), he's also authorized to perform the requested action.

lib/utils/hystrix/src/index.js
```diff
@@ -152,38 +159,42 @@ const command = (s) => {
...
-      // periodically as specified in the client request or every 2000 ms
-      const write = () => {
-        const now = Date.now();
-        const all = values(perf.all(now));
-        if(all.length)
-          map(all, (s) => {
-            res.write('data: ' + JSON.stringify(command(s)) + '\n\n');
-            if(res.flush) res.flush();
+    if(req.path === '/hystrix.stream')
+      oauth.basicAuth(uris.auth_server, req.headers &&
```
jsdelfino: Same comments as for the Eureka module

lib/utils/eureka/src/index.js
jsdelfino: Authentication and Authorization should be optional here, see how we use a SECURED env variable for this everywhere else

lib/utils/oauth/src/index.js
```diff
@@ -278,8 +283,82 @@ const cache = (authServer, id, secret, scopes) => {
...
+        header: {
+          'WWW-Authenticate': 'Bearer realm="cf", error="invalid_token",' +
+            ' error_description="malformed"'
+        }
+      }, false);
+    }
+    const id = decoded[0];
+    const secret = decoded[1];
+    // Give access only to user with the right username and password
+    const scope = id === username && secret === password ?
```
jsdelfino: Not sure why you need to check for this username and password here in the client library as UAA is responsible to authenticate the user for you.

lib/utils/oauth/src/index.js
```diff
@@ -16,12 +16,17 @@ const isEmpty = _.isEmpty;
...
 const intersection = _.intersection;
 const difference = _.difference;
+const contains = _.contains;
 
 const debug = require('abacus-debug')('abacus-oauth');
 const edebug = require('abacus-debug')('e-abacus-oauth');
 
 const get = retry(request.get);
 
+const username = process.env.USERNAME || 'abacus';
```
jsdelfino: Not sure why you need these hardcoded username and password.

KRuelY:
@jsdelfino Hi, I made changes based on feedback. Can you take a look? Thanks for the feedback!

jsdelfino:
Hey @KRuelY would it be possible for you to combine these two commits into one? It's a bit difficult to review right now as the second commit changes code from the first commit. Also merging this pull request now, would result in the first commit (which had issues) still appearing in our commit history.

Thanks a lot!

KRuelY:
Ok, I'll close this and submit a new PR. Thanks!

PR #287, opened 2016-03-30
Secure healthcheck and hystrix stream with basic authentication.

KRuelY:


jsdelfino:
Looks very nice now! Thanks! Merging...

PR #292, opened 2016-04-10
Business responses are returned as an object.

KRuelY:
This changes include changes to collector, metering, rating, and pricing configs.
The summary of the changes is that instead of returning a value, we're returning a business response object.

For example:
Getting metering plan id: Instead of returning 'sample_plan_id' we are returning 
```javascript
{
  error: undefined,
  body: 'sample_plan_id',
  reason: undefined
}
```

In the case that there is an error, instead of throwing an error, we are returning:
```javascript
{
  error: true,
  body: undefined,
  reason: 'metering plan id not found'
}
```

Collector will go through each business responses and make sure that there is no error. When there are error(s), collector will add two fields to the usage doc:
```javascript
{
  error: true,
  reasons: ['account not found', 'metering plan id not found']
}
```

The reason for adding the two fields is so that dataflow would be able to handle the business errors that happen (example: redirect docs with error to an error db). The changes to dataflow is not yet there, but I want to get feedback before making the changes to the dataflow, hence the pull request.

Any feedback would be appreciated. Thanks!

### DIFF
--- a/lib/config/metering/src/index.js
+++ b/lib/config/metering/src/index.js
@@ -107,10 +107,9 @@ const cacheId = (k, id) => {
 const cachedId = (k) => ids.get(k);
 
 // Business response object
-const businessResponse = (error, response) => ({
-  error: error,
-  body: !error ? response : undefined,
-  reason: error ? response : undefined
+const businessError = (reason) => ({
+  error: true,
+  reason: reason
 });
 
 // Retrieve the metering plan id applicable to the specified organization,
@@ -121,7 +120,7 @@ const id = (oid, rtype, ppid, time, auth, cb) => {
   if (!rtype) {
     debug('Undefined resource type');
     return cb(undefined,
-      businessResponse(true, 'Metering config - Undefined resource type'));
+      businessError('Metering config - Undefined resource type'));
   }
 
   // Round time to a 10 min boundary
@@ -132,13 +131,16 @@ const id = (oid, rtype, ppid, time, auth, cb) => {
   const c = cachedId(k);
   if(c) {
     debug('Metering plan id for %s found in cache', k);
-    return cb(undefined, businessResponse(undefined, c));
+    return cb(undefined, {
+      metering_plan_id: c
+    });
   }
 
   // Lookup and cache test plans
   if(rtype === 'test-resource')
-    return cb(undefined, businessResponse(undefined, cacheId(k,
-      require('./test/test-metering-plan').plan_id)));
+    return cb(undefined, {
+      metering_plan_id: cacheId(k, require('./test/test-metering-plan').plan_id)
+    });
 
   // Forward authorization header field to provisioning plugin
   const o = auth ? { headers: { authorization: auth } } : {};
@@ -164,12 +166,14 @@ const id = (oid, rtype, ppid, time, auth, cb) => {
           val.statusCode, val.body || '');
 
         // Response back with description of the error.
-        return cb(undefined, businessResponse(true,
+        return cb(undefined, businessError(
           'Metering config - Unable to retrieve metering plan id'));
       }
 
       // return the metering plan id
-      return cb(undefined, businessResponse(undefined, cacheId(k, val.body)));
+      return cb(undefined, {
+        metering_plan_id: cacheId(k, val.body)
+      });
     });
 };
 
--- a/lib/config/metering/src/index.js
+++ b/lib/config/metering/src/index.js
@@ -106,10 +106,23 @@ const cacheId = (k, id) => {
 // Return a metering plan from the cache
 const cachedId = (k) => ids.get(k);
 
+// Business response object
+const businessResponse = (error, response) => ({
+  error: error,
+  body: !error ? response : undefined,
+  reason: error ? response : undefined
+});
+
 // Retrieve the metering plan id applicable to the specified organization,
 // resource type, provisioning plan and time
 const id = (oid, rtype, ppid, time, auth, cb) => {
   debug('Retrieving metering plan_id');
+
+  if (!rtype) {
+    debug('Undefined resource type');
+    return cb(undefined,
+      businessResponse(true, 'Metering config - Undefined resource type'));
+  }
 
   // Round time to a 10 min boundary
   const t = Math.floor(time / 600000) * 600000;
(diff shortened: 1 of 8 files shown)


jsdelfino:
Hey Kevin,

I think this is going in the right direction but I'm a bit unsure about the new body field you've introduced, as (1) this seems to complicate the valid response case (as the business data is now under that new body field), and (2) unless I mis-understood the change we now have a body inside of the normal HTTP response body, which could become really confusing in my opinion.

How about flattening the data structure a bit and having the error and reason fields inside the business data, instead of putting the business data inside a business response wrapper under a body field?

That would give us the following:
```
// OK case
{
  metering_plan_id: 1234
}

// Error case
{
  error: true,
  reason: 'metering plan mis-configured or whatever',
  // Doesn't even need to be there  
  metering_plan_id: undefined
}
```

Thoughts?

KRuelY:
My original implementation was like what you suggested. I think the reason why I changed it to use body was because I was thinking of putting them in separate module, so that I don't have to construct the business object every time. I agree that 'body' is a bad name because it is similar to http response though. I'll make the changes.

Thanks!

lib/config/metering/src/index.js
```diff
@@ -121,7 +120,7 @@ const id = (oid, rtype, ppid, time, auth, cb) => {
   if (!rtype) {
     debug('Undefined resource type');
     return cb(undefined,
-      businessResponse(true, 'Metering config - Undefined resource type'));
+      businessError('Metering config - Undefined resource type'));
```
jsdelfino: This could also be named simply ```error```, as that'd also be more consistent with the ```error``` field it returns.

lib/config/pricing/src/index.js
```diff
@@ -148,7 +151,7 @@ const id = (oid, rtype, ppid, time, auth, cb) => {
   if(!rtype) {
     debug('Undefined resource type');
     return cb(undefined,
-      businessError(true, 'Pricing config - Undefined resource type'))
+      businessError('Pricing config - Undefined resource type'))
```
jsdelfino: Looks like you've created a pattern here, where your error messages have two parts:
1. sometimes the service that detects the error (collector), or the type of entity that causes it (pricing config)
2. a message (undefined resource type) that can only be understood in the context of (1)

I'd suggest to change these error messages as follows:
- do not include the service name as that won't be relevant to the business guy looking at the error
- use a simple descriptive sentence
- include in that sentence enough info for the business guy to know what to do with it

Here's an initial attempt at doing that for the above error: Pricing config for resource instance 'abcd-1234-def-6789' couldn't be determined as its resource type 'foobar-type' is not a known resource type.

An alternative could be return much more structured data with the error, something like:
```
{
  resource_instance_id: 'abcd-1234-def-6789',
  error: true,
  reason: 'pricing config could not be determined'
  cause: {
    resource_type: 'foobar-type',
    error: true,
    reason: 'unknown resource_type'
  }
```

The more structured approach would allow you to describe the sequence of events that caused an error, providing useful info to the business guy who will need to resolve the problem.

Thoughts?

jsdelfino:
Looks much simpler now, Thanks!

Would you mind squashing the 5 commits in the pull request together into a single commit?
(sorry for that, but that way we won't pollute the history with commits that had issues in them) 

I also made a few comments on your last commit, but they're more for future work that I'd like to see on top of this pull request. I'll be happy to merge the pull request as-is once you squash the commits together.

Thanks!

lib/config/pricing/src/index.js
KRuelY: It is true that the service name is irrelevant to the business person looking at the error because the endpoint to retrieve the error would have collector's service name on it. 

The reason why Undefined resource type is a bad error description is because there is a cascading effect. In the collector, the steps to enrich the usage doc is as follows:
1. get the resource type and get account information.
2. get metering plan id, rating plan id, and pricing plan id.
3. get the price.

Step 2 (get metering plan id, rating plan id, and pricing plan id) would automatically returns business error when step 1 (get resource type) fails . Step 3 (get price) would automatically returns business error when step 2 (pricing plan id) or step 1 (account information) fail.

Which made me question if I should stop the enriching process once an error is found in any of the steps because currently the errors are compiled at the end of steps 3. 

In this pull request, when there are errors when getting account information the collector would post:
```javascript
{
  start: 1420243200000,
  end: 1420245000000,
  organization_id: 'a3d7fe4d-3cb1-4cc3-a831-ffe98e20cf27',
  space_id: 'aaeae239-f3f8-483c-9dd0-de5d41c38b6a',
  consumer_id: 'external:bbeae239-f3f8-483c-9dd0-de6781c38bab',
  resource_id: 'test-resource',
  plan_id: 'basic',
  resource_instance_id: '0b39fa70-a65f-4183-bae8-385633ca5c87',
  measured_usage: [{
    measure: 'light_api_calls',
    quantity: 12
  }],
  error: true,
  reason: [
    'Unable to retrieve account information',
    'Undefined pricing country'
  ]
}
```
Undefined pricing country is the result of unable to retrieve account information.

KRuelY:
Squashed in #298, closing this. Thanks for the feedback!

lib/config/pricing/src/index.js
jsdelfino: Interesting, so maybe there are two different configurations here:

- A flow with 3 steps, step 1 succeeds, step 2 fails, step 3 fails too because step 2 failed to obtain the required data for it. For this case, I'd suggest to stop the execution of the flow in step 2. The reported error is simply the error that caused step 2 to fail.

- Nested calls, with service 1 calling service 2 calling service 3, service 3 fails and reports an error, causing service 2 to fail and report its own error, causing service 1 to fail and report its own error as well. In this case we could either propagate and report the error from service 3 as is, or report the nested errors like I was trying to describe in my previous comment, looking like a sort of 'business error stack trace'. I'm not sure yet about useful that'd be so it's probably better to keep it simple for now and simply propagate the original error from service 3 as is.

What do you think?

KRuelY: I agree with the first point where we should stop the execution of the flow once there is an error in step 1 / 2. 

For the second point: in the pull other pull request that I want to get reviewed, I posted a similar question. I only made the changes in the mapper mainly because it doesn't pick up previous accumulated/aggregated usage and adds more usage on top of it.

In the case of reducer, if aggregator fails, aggregator needs to tell accumulator not to post the updated accumulated usage to the output db. Although I'm not too sure what kind of business error can happen in aggregator, but not in accumulator (Both are grabbing the same metering/rating plan).

Thanks. I'll made changes based on the first point for now!

PR #294, opened 2016-04-11
Added fail event management to dataflow's mapper.

KRuelY:
The dataflow's mapper will now redirects docs with 
``` javascript
{
  error: true
}
```
to the error db instead of to the sink and output db.

For discussion:

Identifying errors. Currently the way to identify business error is to make the map / reduce function attach error field.

The configuration of the dataflow mapper becomes: [test](https://github.com/cloudfoundry-incubator/cf-abacus/blob/76d6c7fa19249ab037cc9a52213942da1220e693/lib/utils/dataflow/src/test/test.js#L77-L106). 

The error doc's id is set to [idoc id/doc order](https://github.com/cloudfoundry-incubator/cf-abacus/blob/76d6c7fa19249ab037cc9a52213942da1220e693/lib/utils/dataflow/src/index.js#L290-L295). I'm not too sure how to get the error doc. I am currently setting it to the input doc id.

The error doc's get endpoint is getting [all error docs from beginning of the month to current time](https://github.com/cloudfoundry-incubator/cf-abacus/blob/76d6c7fa19249ab037cc9a52213942da1220e693/lib/utils/dataflow/src/index.js#L556-L576) in descending order. As an operator I want to see list of errors, which is why I decided to grab all errors.

Enhancement:
1. Add more endpoints that will allow operator to do more operations: example delete a doc after the operator fixes the issue, resubmit, or delete all.
2. Add fail event management to dataflow's groupReduce. I'm hesitant to add the error redirecting in reduce because let's say that a usage passed accumulator and fails at aggregator we will lose accuracy.

Any feedback would be appreciated. Thanks.

### DIFF
--- a/lib/utils/dataflow/src/index.js
+++ b/lib/utils/dataflow/src/index.js
@@ -280,11 +280,17 @@ const postError = (oid, res) => {
 // Build a list of output docs
 const buildOutputs = (itype, idoc, otype, odocs, okeys, otimes, now) => {
   return map(odocs, (odoc, i, l) => {
-    return extend({},
+    // Return doc or error doc
+    return !odoc.error ? extend({},
       odoc,
       idoc.id ? object([[idname(itype), idoc.id]]) : {}, {
         id: dbclient.kturi(okeys[i], otimes[i]),
         processed_id: seqid.pad16(seqid()),
+        processed: now
+      }) : extend({},
+      odoc,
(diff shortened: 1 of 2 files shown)


jsdelfino:
Will review this one again after we settle pull request #292

KRuelY:
Added bunch of enhancement after some thinking. Closing this and continuing in #303

PR #299, opened 2016-04-15
Stop flow when business errors are found in the collector

KRuelY:
In the collector, the steps to enrich the usage doc are as follows:
1. get the resource type and get account information.
2. get metering plan id, rating plan id, and pricing plan id.
3. get the price.

Step 2 (get metering plan id, rating plan id, and pricing plan id) would automatically returns business error when step 1 (get resource type) fails . Step 3 (get price) would automatically returns business error when step 2 (pricing plan id) or step 1 (account information) fail.

The changes in this pull request made it so that the execution of the flow is stopped when the pre-req step fails. The reported error is simply the error that cause the prereq fails.

### DIFF
--- a/lib/config/pricing/src/index.js
+++ b/lib/config/pricing/src/index.js
@@ -69,13 +69,6 @@ const error = (reason) => ({
 const plan = (ppid, country, auth, cb) => {
   debug('Retrieving pricing plan %s', ppid);
 
-  if(!ppid || !country) {
-    debug('Undefined pricing plan id / country');
-    let eMessage = !ppid ? 'Undefined pricing plan id' : '';
-    eMessage += !country ? ' Undefined pricing country' : '';
-    return cb(undefined, error(eMessage));
-  }
-
   return lock(ppid, (err, unlock) => {
     if(err)
       return unlock(cb(err));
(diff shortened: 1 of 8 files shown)


lib/config/pricing/src/index.js
```diff
@@ -69,13 +69,6 @@ const error = (reason) => ({
 const plan = (ppid, country, auth, cb) => {
   debug('Retrieving pricing plan %s', ppid);
 
-  if(!ppid || !country) {
-    debug('Undefined pricing plan id / country');
-    let eMessage = !ppid ? 'Undefined pricing plan id' : '';
```
jsdelfino: Not sure we need that mutation of eMessage here

lib/config/pricing/src/index.js
```diff
@@ -69,13 +69,6 @@ const error = (reason) => ({
...
-    eMessage += !country ? ' Undefined pricing country' : '';
```
jsdelfino: I'd suggest to have two different error messages here (and two different returns) instead of trying to combine the two (which could be another possible approach but then we'd need to generalize it to all the other cases as well).

KRuelY: Completely missed that. Thanks for catching it! 

With the collector's flow changes, ppid and country would always be defined.

Will remove that and stash the commit.

Issue #239, opened 2016-02-16
zip not specified as a requirement


prysmakou:
Please mention zip as a requirement.

jsdelfino:
Thanks for reporting this.

I think it'd be nice to actually remove that dependency on the zip executable and instead change our cfpack tool to use one of the many Node.js zip modules available out there.

Thoughts?

hsiliev:
With the removal of cfpack we now no longer use zip

PR #301, opened 2016-04-19
Remove Zip Dependency

rajkiranrbala:
Addresses issue https://github.com/cloudfoundry-incubator/cf-abacus/issues/239

### DIFF
--- a/tools/cfpack/src/index.js
+++ b/tools/cfpack/src/index.js
@@ -8,8 +8,9 @@
 var _ = require('underscore');
 var fs = require('fs');
 var path = require('path');
-var cp = require('child_process');
 var commander = require('commander');
+var archiver = require('archiver');
+var globby = require('globby');
 
 var map = _.map;
 var pairs = _.pairs;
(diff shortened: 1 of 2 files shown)


tools/cfpack/src/index.js
```diff
@@ -54,27 +55,40 @@ var repackage = function(root, cb) {
...
 // Produce the packaged app zip
-var zip = function(ignore, cb) {
+var zip = function(cb) {
   fs.unlink(path.resolve('.cfpack', 'app.zip'), function(err) {
+
     if(err) noop();
 
-    // We're using the system zip command here, may be better to use a
-    // Javascript zip library instead
-    var files = '-type f -not -regex "\\./\\.cfpack/package\\.json" ' +
```
jsdelfino: Is the new code still excluding .cfpack/package.json and .git?

jsdelfino:
I applied your patch to my local dev env and ```npm run cfpack``` in ```lib/metering/collector``` never completes. Are you running into the same issue?

rajkiranrbala:
If the .cfpack directory contains soft links that do no match excluded patterns then it goes into an infinite loop. Try to fix this in a different way

PR #303, opened 2016-04-20
Added fail event management to dataflow

KRuelY:
When the dataflow's mapper and reduce function returns:
``` javascript
{
  error: true
}
```
it will redirect the doc to the error db instead of to the sink and output db.

In summary:
1. When the sink returns error, in addition to the statusCode 201 the current service will returns a body with error and reasons without posting any of its docs to the output db.
2. When the map / reduce functions produces error, in addition to returning statusCode 201, the current service will returns a body with the error and reasons without posting any of its docs to the output db / to the sink. Instead it will post to error db.

For discussion:
1. The structure of the error object: (currently it is just a string.)
2. When the sink returns error, should the current service also post to its error db? (Each service has its own db)
3. An error dedicated service to manage errors instead of each service has its own db. (The set up would be similar to the sink. Each service will post to the error dedicated service instead of its own db)

Enhancement:
1. Add more endpoints that will allow operator to do more operations: example delete a doc after the operator fixes the issue, resubmit, or delete all.

Any feedback would be appreciated. Thanks.

### DIFF
--- a/lib/utils/dataflow/src/index.js
+++ b/lib/utils/dataflow/src/index.js
@@ -27,6 +27,7 @@ const extend = _.extend;
 const map = _.map;
 const object = _.object;
 const filter = _.filter;
+const find = _.find;
 const pairs = _.pairs;
 const zip = _.zip;
 const last = _.last;
@@ -280,11 +281,17 @@ const postError = (oid, res) => {
 // Build a list of output docs
 const buildOutputs = (itype, idoc, otype, odocs, okeys, otimes, now) => {
   return map(odocs, (odoc, i, l) => {
-    return extend({},
+    // Return doc or error doc
+    return !odoc.error ? extend({},
       odoc,
       idoc.id ? object([[idname(itype), idoc.id]]) : {}, {
         id: dbclient.kturi(okeys[i], otimes[i]),
         processed_id: seqid.pad16(seqid()),
+        processed: now
+      }) : extend({},
+      odoc,
+      idoc.id ? object([[idname(itype), idoc.id]]) : {}, {
+        id: [idoc.id, i].join('/'),
         processed: now
       });
   });
@@ -311,6 +318,7 @@ const postOutput = function *(odoc,
       throw postError(odoc.id, res);
 
     debug('Posted %s successfully to sink', odoc.id);
+    return res;
   }
   catch(exc) {
     edebug('Exception posting %s to sink, %o', odoc.id, exc);
@@ -319,12 +327,27 @@ const postOutput = function *(odoc,
   }
 };
 
+// Check if any of the docs have errors in it
+const validateDocs = (docs) => find(docs, (doc) => doc.error) ? [] : docs;
+
 // Post a list of output docs to the configured sink service
 const postOutputs = function *(odocs,
   shost, spartition, spost, authentication) {
-  yield tmap(odocs, function *(odoc, i, l) {
-    return yield postOutput(odoc, shost, spartition, spost, authentication);
-  });
+  // Only post when all docs have no error.
+  const reasons = yield treduce(validateDocs(odocs), function *(a, odoc) {
+    const res = yield postOutput(odoc, shost, spartition, spost,
+      authentication);
+    // accumulate reason of errors from the sink
+    return res.body && res.body.error ? a.concat(res.body.reasons) : a
+  }, []);
+  return reasons;
+};
+
+// Log an error doc
+const logError = function *(edoc, edb) {
+  debug('Logging error doc %s', edoc.id);
+  yield edb.put(edoc);
+  debug('Logged error doc %o', edoc);
 };
 
 // Log an output doc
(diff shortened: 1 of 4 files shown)


jsdelfino:
> For discussion:
> 1. The structure of the error object: (currently it is just a string.)

I'd suggest something similar to how Cloudant returns errors:
- an error string field containing a keyword e.g. einvalidplan, eorgnoutfound
- a reason string field containing a text description of the reason that caused the error  

> 2. When the sink returns error, should the current service also post to its error db? (Each service has its own db)

Yes

> 3. An error dedicated service to manage errors instead of each service has its own db. (The set up would be similar to the sink. Each service will post to the error dedicated service instead of its own db)

Could be useful, but on top of the error db, as it'll be useful for each service to have its own db in distributed topologies, and posting to an error sink service could fail as well... 

> Enhancement:
> 1. Add more endpoints that will allow operator to do more operations: example delete a doc after the operator fixes the issue, resubmit, or delete all.

Yes, makes sense

lib/utils/dataflow/src/index.js
```diff
@@ -280,12 +281,18 @@ const postError = (oid, res) => {
 // Build a list of output docs
 const buildOutputs = (itype, idoc, otype, odocs, okeys, otimes, now) => {
```
jsdelfino: Not sure why we're building different content when error is true, for example why are we not including the processed_id field in that case

lib/utils/dataflow/src/index.js
```diff
@@ -335,11 +358,28 @@ const logOutput = function *(odoc, odb) {
...
 // Log a list of output docs
-const logOutputs = function *(odocs, odb) {
-  yield tmap(odocs.concat([]).reverse(), function *(odoc, i, l) {
-    // Log each doc into the output database
-    yield logOutput(odoc, odb);
-  });
+const logOutputs = function *(odocs, odb, edb) {
+  let errors;
+  // if there is an error, post to error db
+  if(!validateDocs(odocs).length) {
```
jsdelfino: Is this really a validation? Looks like it's only checking for the presence of an error in any doc, and then either returning an empty array or the whole list of docs, so that doesn't look like a validation to me, and also not sure about the purpose of returning an array (empty or the input array) instead of just a true/false.

lib/utils/dataflow/src/index.js
```diff
@@ -523,6 +580,34 @@ const mapper = (mapfn, opt) => {
       };
     }));
 
+  // Retrieve error docs
+  if(opt.error && opt.error.get)
+    routes.get(opt.error.get, throttle(function *(req) {
+      const ts = map(filter(
+        pairs(req.params), (p) => /^t/.test(p[0])), (p) => p[1]).join('/');
+
+      debug('Retrieving error docs from beginning of the month to %s', ts);
```
jsdelfino: Why just the beginning of the month?

lib/utils/dataflow/src/index.js
```diff
@@ -380,9 +432,10 @@ const mapper = (mapfn, opt) => {
 
   // Map an input doc to an output doc, store both the input and output and
   // pass the output to the configured sink service
-  /* eslint complexity: [1, 6] */
+  /* eslint complexity: [1, 10] */
```
jsdelfino: Looks like a workaround to make eslint happy, wouldn't it be better to actually lower the complexity by splitting the code into multiple functions?

lib/utils/dataflow/src/index.js
```diff
@@ -622,14 +710,14 @@ const reducer = (reducefn, opt) => {
   // Configure our batch grouping reduction function
   const greduce = groupReduce(opt.input.type, yreducefn,
     opt.output.type, ocache, odb,
-    opt.sink.host, opt.sink.apps, opt.sink.post);
+    opt.sink.host, opt.sink.apps, opt.sink.post, edb);
```
jsdelfino: I'd suggest to move edb to just after odb, as errors are just an alternate type of output

lib/utils/dataflow/src/index.js
```diff
@@ -319,12 +327,27 @@ const postOutput = function *(odoc,
   }
 };
 
+// Check if any of the docs have errors in it
+const validateDocs = (docs) => find(docs, (doc) => doc.error) ? [] : docs;
+
 // Post a list of output docs to the configured sink service
 const postOutputs = function *(odocs,
   shost, spartition, spost, authentication) {
-  yield tmap(odocs, function *(odoc, i, l) {
```
jsdelfino: Why do we need to change a map to a reduce here? Looks like this will have a significant impact on performance, as we won't be able to parallelize the posts anymore...

lib/utils/dataflow/src/index.js
KRuelY: The reason is because I think the id should be based on the input key rather than the output key.
Instead of setting id = dbclient.kturi(okeys[i], otimes[i]), I set the id of the error doc to be based on the input doc's id, hence id = [idoc.id, i].join('/') where i = the doc index of the input array.

I think the reason why I didn't attach processed_id is because I thought of it as a field to detect duplicate, and I thought that the error doc is an unprocessed doc because it is just the input doc + error and reasons field.

Should we just leave them to be the same?

lib/utils/dataflow/src/index.js
KRuelY: This was the simplest implementation of get that I thought of at the moment.

Operator would pass a time, and dataflow would returns all errors in the database from the beginning of the month(based on the time passed) to the time that the operator passed.

Should I returns all docs? I didn't put much thought into this because I am not sure with the use case, and would like to get actual comments from users and enhance.

lib/utils/dataflow/src/index.js
KRuelY: Yeah you're right. I tried to split the code, but I was unsure on how to split it and I thought that I should get feedback first.

lib/utils/dataflow/src/index.js
KRuelY: +1

lib/utils/dataflow/src/index.js
KRuelY: I actually change this back to tmap in later commits in this pull request(my bad for not stashing them). I was trying out treduce because I thought the doc would just be a single doc everytime after the change in collector, but aggregator proves me wrong.

KRuelY:
Will submit a new pull request after making some clean up and adjustment due to commit f412896. Thanks!

PR #304, opened 2016-04-21
Remove Zip Dependency

rajkiranrbala:
Addresses issue https://github.com/cloudfoundry-incubator/cf-abacus/issues/239

### DIFF
--- a/tools/cfpack/package.json
+++ b/tools/cfpack/package.json
@@ -32,7 +32,9 @@
     "pub": "publish"
   },
   "dependencies": {
+    "archiver": "^1.0.0",
     "commander": "^2.8.1",
+    "shelljs": "^0.6.0",
     "underscore": "^1.8.3"
   },
   "devDependencies": {
--- a/tools/cfpack/src/index.js
+++ b/tools/cfpack/src/index.js
@@ -8,8 +8,9 @@
 var _ = require('underscore');
 var fs = require('fs');
 var path = require('path');
-var cp = require('child_process');
 var commander = require('commander');
+var archiver = require('archiver');
+require('shelljs/global');
 
 var map = _.map;
 var pairs = _.pairs;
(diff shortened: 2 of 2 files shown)


jsdelfino:
Thanks for the update
```npm run cfpack``` in lib/metering/collector gives me the following error:

abacus-usage-collector@0.0.5-rc.1 cfpack /Users/jsdelfino/Desktop/github/cf-abacus/lib/metering/collector
cfpack

events.js:154
      throw er; // Unhandled 'error' event
      ^

Error: EMFILE: too many open files, open 'node_modules/abacus-eslint/node_modules/eslint/lib/rules/indent.js'
    at Error (native)

Can you try as well and see if you are getting the same error? Thanks!

**Comment by rajkiranrbala** [comment]
Instead of individually adding the files, I pass an array of paths to the bulk command which prevents the Error: EMFILE from occurring. Also I am considering only files listed in package.json in addition to contents of .cfpack

**Comment by betafood** [comment]
Thanks for the update! I was able to produce the zip on my environment with your newest commits. There's one thing I wanted to address from the resulting zip.

It looks like the ignore parameter in the zip function is no longer being used, so the contents of the zip are now different.
To give an example, within each component or module(like aggregator), their respective lib directory is being packed when previously it was not. This is same situation for other ignored patterns like .log files. Ideally, I think we would probably want the same resulting zip with this dependency change. Thoughts?

**Comment by KRuelY** [comment]
Hi thanks for the pull request. I was able to produce the zip on lib/metering/collector.
I don't see the problems mentioned by the comments by @betafood and @jsdelfino.

Here are the two zip produced by the master and the pull request:
[masterapp.zip](https://github.com/cloudfoundry-incubator/cf-abacus/files/246179/masterapp.zip)
[pullrrequestapp.zip](https://github.com/cloudfoundry-incubator/cf-abacus/files/246180/pullrrequestapp.zip)

I run the cfpack with npm-debug.log and I don't see the log in the zip.

**Comment by betafood** [comment]
You're correct. I had not yet run the build in my master branch test, so it does pack the lower lib directories.

However, I also did try having an npm-debug.log file inside of aggregator, and it was packed inside the zip. My first statement with the regards to the lib directories is invalid, but other patterns in previously ignored may not be.

For the most part, the ignored patterns are defined by the .gitignore file (in the default case), so I believe those may still need to be addressed here.

**Comment by KRuelY** [comment]
Some files are not ignored by this pull request. Here are the steps to replicate:
1. Build
2. Go to lib/metering/collector and generate a log file (ex: npm run blabla)
3. Go to tools/cfpack and run these commands.
```shell
./cfpack
cd .cfpack
unzip app.js -d app
cd app/.cfpack
cd lib/metering/collector
```
The log file generated in step 2 would not be ignored.

rajkiranrbala:
The original signature of the zip function was accepting a ignore parameter. The only place were the function is called is at [Line 136](https://github.com/cloudfoundry-incubator/cf-abacus/blob/d23f8ec3d6cbb44a07bf7dea958739fd4711cf07/tools/cfpack/src/index.js#L136) and the only argument passed is the path of .gitignore. Now that the .gitignore is ignored as a part of the filter list, I removed the ignore attribute from the function.

The function doesn't ignore any other file other than the ones specified in the list.

jsdelfino:
Any update on this? Has anybody validated that this latest pull request addresses the issues previously reported with it, and is it ready to be merged?

jsdelfino:
@rajkiranrbala Any update on this pull request? Is it ready to be merged?

rajkiranrbala:
@jsdelfino 
The contents of the package are slightly different, but it doesn't affect the application running in cloud foundry.

rajkiranrbala:
I am closing this pull request. We need to find a way to mimic the exact behavior of linux zip command.

PR #307, opened 2016-04-25
Rating and metering plan returns business response.

KRuelY:
Summary:

- Metering & rating config's get plan function returns business response.
- Adjusted Meter, Accumulator, and Aggregator to attach error and reasons to doc when error happen.
- Report will throw error when business error happens.


KRuelY:
Combined and commited in f412896

PR #331, opened 2016-06-06
Introduce metrics type in metering plans

hsiliev:
Added mandatory `type` string field to the metering plans


KRuelY:
The metering/rating/pricing schema check only happen at the provisioning plugin to allow the integrator to use their own schema and add additional fields that their custom provisioning plugin need to resolve metering/rating/pricing plan. Adding the metering metrics type here as required will not break any implementation that we have until we use the type field outside of the plugins.

PR #338, opened 2016-06-15
Pass the timewindow with cell function to the accumulate and aggregate.

KRuelY:
The goal is to add more flexibility in the formula by giving the formula access to the hourly/minutely/daily accumulated / aggregated usage.

An example of use case is if we are in the month window, and we are interested to see the daily usages. In order to do so, the formula need to access the timewindow.

I also added a .cell method in the timewindow to let the formula knows which window the usage it is calculating belongs to.

### DIFF
--- a/lib/aggregation/accumulator/src/index.js
+++ b/lib/aggregation/accumulator/src/index.js
@@ -196,6 +196,15 @@ const accumulate = function *(a, u) {
 
       const aw = accum(amerge.accumulated_usage, mu.metric);
 
+      // a function that gives the index of the usage in the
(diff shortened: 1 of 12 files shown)


KRuelY:
After discussion with @jsdelfino, the timewindow is not passed to the accumulation or aggregation anymore. Instead a function(takes in dimension: 's', 'm', 'h', 'D', 'M') to get the cell in which the current usage belongs to in the time window dimension is passed. This is so that we can avoid modifying the timewindow from the formula.

I also I moved the cell function to the timewindow module.

Issue #336, opened 2016-06-14
Dataflow module does not replay any records.


rajkiranrbala:
The data flow module has a replay function that will not be able to replay any docs. With the range query defined in [L1061-L1063](https://github.com/cloudfoundry-incubator/cf-abacus/blob/e2fe32950e6bcd5f300391b2dfdbc8c6f2d360a5/lib/utils/dataflow/src/index.js#L1061-L1063) there will be no documents falling within the range of start key and end key.

For example the id in the database for the input doc would look like 
`t/0001465926293403-0-0-1-0/k/anonymous`

The start key and the end key  for the range query would like
`t/0001465926200000/k/`
`t/0001465926300000/k/ZZZ`

In order to retrieve the documents within the range, we must be able to do the following 

1. Run the query against a specific bucket (Since there is a one to one relationship between the instance running the replay function and the db partition bucket)
2. The start key and end key should include only time and no key.

In order do the above two, we need the following changes

1. ~~The replay function must know what partition bucket(instance id) it needs to work on~~
2. The dbclient must be able work with start  and end keys that include only time in case of (tk uris)
3. ~~The partitioning functioning in the replay logic should be able to direct its requests against a specific partition bucket.~~

Edit

Items 1 and 3 are already being done in the dataflow module.

hsiliev:
Tried to solve the same with the #88: https://github.com/hsiliev/cf-abacus/blob/compensation/lib/cf/compensation/src/index.js#L177-L192

rajkiranrbala:
@hsiliev 

I am thinking about customizing the forward function that is passed to the partitioner. If the bucket is undefined then it should return all the possible combinations of partitions for a given time range

KRuelY:
Looks like the problem is in dbclient's range queries. We had issues regarding this before, and it is probably a good idea to enhance dbclient.

https://github.com/cloudfoundry-incubator/cf-abacus/issues/308
https://github.com/cloudfoundry-incubator/cf-abacus/issues/34

rajkiranrbala:
I was trying to solve this by modifying the forward function to return all the partitions for a given time range. However the balance function returns only one partition. I am going to modify the couchclient so that it can work on an array of partitions

rajkiranrbala:
Fixed with the following changes

Changes three modules

#### abacus-partition 

[...]

PR #339, opened 2016-06-19
Data flow replay

rajkiranrbala:
Addresses issue https://github.com/cloudfoundry-incubator/cf-abacus/issues/336

### DIFF
--- a/lib/utils/couchclient/src/index.js
+++ b/lib/utils/couchclient/src/index.js
@@ -20,6 +20,8 @@ const extend = _.extend;
 const zip = _.zip;
 const defaults = _.defaults;
 const omit = _.omit;
+const first = _.first;
+const rest = _.rest;
 
 // Setup debug log
 const debug = require('abacus-debug')('abacus-couchclient');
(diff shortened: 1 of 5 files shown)


rajkiranrbala:
Changes three modules

#### abacus-partition 

- Modified the partitioner function so that it accepts an additional parameter `checkkey`. When `checkkey` is `true` then the balancing is skipped when the key is undefined for a `read` operation
- Bucket functions return `NaN` if the key is undefined
- Forward function returns all possible partitions for the given time period when key is undefined for a `read` operation and throws an error for `write` operation

#### abacus-couchclient

- For a range operation when key is undefined or the startkey and endkey have different keys then pass key as undefined to the partition function
- If the partition function returns an array of unblanced partition read through all the partitions

#### abacus-dataflow

- Does not include key in the range query for finding records for replaying

hsiliev:
We have to add the changes from abacus-couchclient to mongoclient as well. Created issue https://github.com/cloudfoundry-incubator/cf-abacus/issues/340 for this.

PR #348, opened 2016-06-27
Support doc updates on MongoDB

### LINKED ISSUE #347: Dataflow tests fail on MongoDB
The tests fail with:
```
Testing...

  abacus-dataflow
    dataflow map
      ✓ runs a map transform and stores its inputs and outputs (909ms)
      ✓ does not store map outputs when error exists (829ms)
      ✓ propagates mapper error from the sink and does not store to output (915ms)
      ✓ stores map errors to error db (888ms)
      ✓ removes mapper error docs from db (306ms)
    dataflow reducer
      ✓ runs a reduce transform and stores its inputs and outputs (806ms)
      1) runs a reduce transform and down samples outputs
      ✓ does not store reduce outputs when error exists (457ms)
      ✓ propagates reducer error from the sink and does not store to output (560ms)
      ✓ does batch post, and gather the errors (256ms)
      ✓ stores reducer errors to error db (562ms)
      ✓ removes reducer errors docs from db (258ms)


  11 passing (7s)
  1 failing

  1) abacus-dataflow dataflow reducer runs a reduce transform and down samples outputs:

      Uncaught AssertionError: expected 409 to equal 201
      + expected - actual

      -409
      +201

      at lib/test/test.js:1444:40
      at Request._callback (/Users/development/workspace/cf-abacus/node_modules/abacus-request/lib/index.js:167:14)
      at Request.self.callback (/Users/development/workspace/cf-abacus/node_modules/request/request.js:200:22)
      at Request.<anonymous> (/Users/development/workspace/cf-abacus/node_modules/request/request.js:1067:10)
      at IncomingMessage.<anonymous> (/Users/development/workspace/cf-abacus/node_modules/request/request.js:988:12)
      at endReadableNT (_stream_readable.js:913:12)
      at _combinedTickCallback (internal/process/next_tick.js:74:11)
      at process._tickCallback (internal/process/next_tick.js:98:9)
```

jsdelfino:
Adjust the mongoclient module to provide support for updates, _rev and
rev fields compatible with the couchclient module.

Put now uses updateOne if the given doc contains a _rev field.

Batch_put and bulkDocs now use a MongoDB Bulk op with either updateOnes
for updates and insertOnes for inserts.

Documents returned by geti, batch_get, and allDocs now return a _rev field
set to '1' to allow the caller to detect that the doc comes from a version
of the doc in the DB (as opposed to a new doc that's not been written to
the DB yet).

Put, batch_put, remove, batch_remove and bulkDocs now return a rev field
set to '1' in their results, to allow the caller to use that rev field
and propagate it to updates of the retrieved doc.

Fixes #347.

### DIFF
--- a/lib/utils/mongoclient/src/index.js
+++ b/lib/utils/mongoclient/src/index.js
@@ -193,7 +193,7 @@ const errdb = (name, err) => ({
     deleteOne: (doc, cb) => setImmediate(() => cb(err)),
     deleteMany: (docs, opt, cb) => setImmediate(() => cb(err)),
     initializeOrderedBulkOp: () => ({
-      insertOne: (doc) => {},
+      insert: (doc) => {},
       find: (id) => ({
         upsert: () => ({
           updateOne: (doc) => {}
@@ -576,7 +576,7 @@ const mongoclient = (part, uri, cons) => {
             cb(null, undefined);
             return;
           }
-          cb(error(err), val);
+          cb(error(err), val ? extend(val, { _rev: '1' }) : val);
         });
       }, 'read', {
         _id: id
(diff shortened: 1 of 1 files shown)


maximilien:
Interesting there seems to be no tests changes for this? Is that correct?

jsdelfino:
IIRC these changes fixed a break of an existing test, see #347, so we didn't want to change the test here, we wanted to fix a bug uncovered by that test (which was not detected early because at that point in time we had no Travis build with MongoDB... we have one now so hoping to not run into this kind of issue anymore...) HTH

maximilien:
Thanks for the background @jsdelfino

PR #359, opened 2016-07-05
Use separate keys to address sink partitions

jsdelfino:
This is for @rajkiranrbala to review and see if that'll address the concurrent issue we discussed today (I asked @rajkiranrbala to create a Github issue for this too).


jsdelfino:
Merged in #363.

Issue #364, opened 2016-07-07
Duplicate Detection while extending the Pipeline


rajkiranrbala:
For the apps that follow the aggregator, currently there is no fool proof way to detect duplicates. If the duplicate detection doc id can be included as a part of aggregated docs, then we can use that information to detect duplicates while extending the pipeline.

rajkiranrbala:
@jsdelfino and I discussed yesterday and came to a conclusion that we should include start, end, consumer_id, resource_instance_id and organization_id that in the org aggregated documents, so that the information can be used to identify the duplicates for apps that are being used to extend the pipeline,

jsdelfino:
Fixed by PR #365

PR #365, opened 2016-07-08
Duplicate Detection while extending the Pipeline #364

rajkiranrbala:
Includes resource_instance_id and consumer_id as a part of org aggregated document. Addresses issue https://github.com/cloudfoundry-incubator/cf-abacus/issues/364

### DIFF
--- a/lib/aggregation/aggregator/src/index.js
+++ b/lib/aggregation/aggregator/src/index.js
@@ -410,16 +410,27 @@ const aggregate = function *(aggrs, u) {
   const newa = a ?
     extend(reviveOrg(JSON.parse(JSON.stringify(a))), {
       account_id: u.account_id,
-      end: u.end
-    }) :
-    extend(newOrg(u.organization_id), {
       start: u.start,
       end: u.end,
-      account_id: u.account_id
+      resource_instance_id: u.resource_instance_id,
+      consumer_id: u.consumer_id
+    }) :
+    extend(newOrg(u.organization_id), {
+      account_id: u.account_id,
+      start: u.start,
+      end: u.end,
+      resource_instance_id: u.resource_instance_id,
+      consumer_id: u.consumer_id
     });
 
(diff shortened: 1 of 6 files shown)


test/aggregation/reporting/src/test/test.js
```diff
@@ -562,6 +563,8 @@ describe('abacus-usage-reporting-itest', () => {
       id: dbclient.kturi(oid(o), end + u),
       organization_id: oid(o),
       account_id: '1234',
+      resource_instance_id: 'rid',
+      cid: 'cid',
```
jsdelfino: Shouldn't this be ```consumer_id: 'cid'```?

lib/aggregation/aggregator/src/index.js
```diff
@@ -410,12 +410,17 @@ const aggregate = function *(aggrs, u) {
...
+      end: u.end,
+      resource_instance_id: u.resource_instance_id,
+      consumer_id: u.consumer_id
     }) :
     extend(newOrg(u.organization_id), {
+      account_id: u.account_id,
       start: u.start,
       end: u.end,
-      account_id: u.account_id
+      resource_instance_id: u.resource_instance_id,
```
jsdelfino: Shouldn't the ```start```, ```end``` and ```resource_instance_id``` fields be added to the consumer level aggregated usage doc as well? If the intent is to help a downstream app do dup detection, then AIUI an app interested in consumer level aggregation would need these fields too in that doc, right?

jsdelfino:
The latest changes in this pull request break the build (see above) and I'm seeing a build break when building locally as well. Any idea?

hsiliev:
As far as I can see the Travis build was failing due to `cf-bridge` and `cf-renewer` tests.

I created https://github.com/cloudfoundry-incubator/cf-abacus/issues/366 to investigate and fix this.

rajkiranrbala:
@jsdelfino Issue #336 was fixed by @hsiliev and the Travis build succeeds after that.

PR #379, opened 2016-08-05
changed eureka's dependency on SECURED for performance

ArchanPatel890:
This allows the secured flag to be set for eureka's healthcheck with the SECURED_HEALTH flag. This allows the rest of the modules to be secure will letting eureka stay open so I can handle a large amount of requests without being dependent on outside tokens.


rajkiranrbala:
@ArchanPatel890 

1. IMO the default value of SECURED_HEALTH should be same as that of SECURED. If SECURED_HEALTH is not set then follow the SECURED variable value. This will help to retain the security level in environments running Abacus without any change in environment variables.

2. The variable name SECURED_HEALTH seems a little hard to understand. How about SECURE_HEALTHCHECK or HEALTHCHECK_SECURED?

Issue #378, opened 2016-08-01
Resource level aggregation usage.


KRuelY:
In aggregation we have resource level aggregated usage and plan level aggregated usage.
The resource level aggregated usage restrict us from having a resource with variety of plans.

Example:
  resource quantity: 9 (number)
  plan A quantity: 5 (number) (Addition: old usage + new incoming usage)
  plan B quantity: 4 (number) (Addition: old usage + new incoming usage)

But the pipeline will throw an error if the two plans' aggregation formula results in different object structure.

Example:
  resource quantity: break..
  plan A quantity: 5 (number) (Addition)
  plan B quantity: { sum: 16, avg: 4, count: 4 } (object) (Average: sum / count)

Solution:
Remove resource level aggregated usage in aggregation. If resource level aggregated usage is needed we can change the logic in reporting app: Instead of using the summarize function in the resource level aggregated usage from aggregator, we sum up all the plans' summary and use that as the resource level aggregated usage. The restriction here is that the summarize function has to output in the same format.

Edit: Adding up all the plans' summary doesn't seems to makes sense too. Using the example above, Adding up the summary of addition with the summary of average doesn't seems right because the two are not related. Would it be better to remove resource level aggregated usage's summary, cost, and quantity? The resource level aggregated usage would only have charge, which will be computed at reporting or would it be better to remove resource level aggregated usage completely?

Thoughts?

KRuelY:
Fixed in #406

PR #380, opened 2016-08-05
Resource aggregated

KRuelY:
Resolve issue #378. The changes here will only keep charge in the resource level aggregated_usage.


KRuelY:
Going to clean up the commits.

PR #425, opened 2016-09-17
Performance fixes 

KRuelY:
This PR contains fixes from me and @rajkiranrbala.

Over the past 1~2 weeks, we've been working on improving the performance and correctness of abacus. The changes in this PR fix most of the problems that we encountered when abacus receive heavy submissions.

**(1) Bucketing based on the first part of the key.**
We want to post all the related documents to the same DB partition. This will improve performance because it allows us to perform batch put instead of individual put to different partition. This will also make debugging a lot easier because all the related documents would be in the same DB partition.

Edit: More than performance, it is  attributed towards correctness. We were seeing in some cases that one of the documents is written to the database while the other isn't. Since we do not have any transactions in NoSQL databases, the best way we could ensure that all the documents are sent to the database at once is by using a batch_put (calling bulk docs in case of Cloudant) (Though this isn't completely guaranteed yet!). In order to do that we need all the documents to be written to the same database. Hence this change was made.

**(2) Streamline the pipeline based on the organization_id & Dataflow batches sink post by sink key.**
During the performance testing, we encountered many 502s. The reason of the 502s are because the some batch requests are blocking other batch requests. This happens when an app receives many batch requests from previous app instances in the pipeline, mixes the received requests, and finally constructs and sends new batch requests to next app instances in the pipeline.

By changing the sink key to organization id and introducing X-BATCH-ID when grouping request, the batches are more focused and we are able to avoid the situation described above.

**(3) Dataflow batch put output docs to the database & Dataflow clean accumulated docs cache when logOutputs fails.**
When doing put to the database, we have to either successfully post the output docs or not have them in the database. With (1), we are able to perform batch put instead of individual put. Individual put causes a problem, when db put timeout, the application will return an error, but it doesn't mean that the aggregated document was not posted successfully to the db. The next submission that comes after would get rejected(409) because the cache contains the old aggregated document, but the db has the newer version of the aggregated document. When the cache refresh / the app restart, we will have incorrect aggregated document because some of the usage was rejected.

Feels free to add more @rajkiranrbala.


lib/utils/dataflow/src/index.js
```diff
@@ -715,11 +735,11 @@ const mapper = (mapfn, opt) => {
       return {
         body: dbclient.undbify(doc)
       };
-    }));
+    });
 
   // Retrieve error docs
   if(opt.error && opt.error.get)
-    routes.get(opt.error.get, throttle(function *(req) {
```
hsiliev: Wouldn't this cause a problem with too much incoming requests? Like out of threads/resources?

Do we rely on the batch calls to the pipeline and the throttling there?

rajkiranrbala: @hsiliev Actually having a throttled request handler causes problem when there are too many incoming requests. The throttled handler doesn't reject any of the incoming requests  when the throttle limit is reached. Instead it just queues all the requests. Though the Go Router log shows 502 after the timeout configured, the request is still living in the queue of the throttle. This prevents the app from processing any of the incoming requests.

hsiliev: Wouldn't it be better if we change throttle to fail the requests then? 

I imagine it can:
* store request and timestamp
* reject/ignore requests older than some time
* can be configured with some timeout

KRuelY:
Removing the bloom filter is necessary because with bloom filter in place, we're getting wrong result.

The last implementation of bloom filters was incorrect because when the app restart, the bloom filters would be empty. All duplicate submissions would be marked as non duplicate, and it would be accumulated to the usage doc, and it would throw 409 due to duplicate when posting the duplicate document to the db.

To fix this error, in this pull request, we removed bloom filters which makes the performance worse because it does a get to the db for every submission to check for duplicate.

With bloom filters removed, the performance take a hit because it did a get for every usage doc to check for duplicate. We will need to find a way to initialize or serialize and deserialize bloom filters when the app restart.

lib/utils/dataflow/src/index.js
rajkiranrbala: @hsiliev 
+1 for that.
Outgoing requests must be throttled and incoming requests must be rate limited.

KRuelY:
Going to close this PR and submit small changes that will not change the current behavior. Will revisit the performance when we pinpoint the problem.

PR #450, opened 2016-11-12
Drop unused renewer client credentials.

jmcarp:
The concourse pipeline specifies renewer client credentials but doesn't appear to use them anywhere, and it looks like the renewer app expects to use the system credentials: https://github.com/cloudfoundry-incubator/cf-abacus/blob/master/lib/cf/renewer/src/index.js#L89-L93. This patch drops the renewer credentials.


hsiliev:
@jmcarp Nice catch

PR #451, opened 2016-11-16
Add error reporting to bridge

svgradev:
Signed-off-by: Stoyan Gradev <stoyan.gradev@sap.com>
Signed-off-by: Plamen Doychev <plamen.doychev@sap.com>

### DIFF
--- a/lib/cf/bridge/src/index.js
+++ b/lib/cf/bridge/src/index.js
@@ -137,6 +137,24 @@ const statistics = {
   }
 };
 
+const errors = {
+  missingToken: false,
+  noReportEverHappened: true,
+  consecutiveFailures: 0,
+  lastError: '',
+  lastErrorTimestamp: ''
+};
+
+const registerError = (message, error, name, time) => {
+  errors.lastError = message;
+  errors.lastErrorTimestamp = new Date().toISOString();
+
+  edebug('%s error: %j', message, error);
+
+  if (name && time && error)
+    perf.report(name, time, undefined, new Error(error));
+};
+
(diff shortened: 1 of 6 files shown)


lib/cf/bridge/src/index.js
```diff
@@ -137,6 +137,25 @@ const statistics = {
...
+  consecutiveFailures: 0,
+  lastError: '',
+  lastErrorTimestamp: ''
+};
+
+const registerError = (message, error, name, time) => {
+  errors.lastError = message;
+  errors.lastErrorTimestamp = new Date().toISOString();
+
+  const msg = util.format('messase: %j error: %j', message, error);
```
hsiliev: I believe we can use the format functionality of the debug module directly and use:
```
edebug('%s error: %j');
```

lib/cf/bridge/src/index.js
```diff
@@ -305,11 +326,14 @@ const reportUsage = (usage, token, config, cb = () => {}) => {
       cb(error, response);
       return;
     }
-    const message = util.format('Failed reporting\n\tusage %j\n\terror %j' +
-      '\n\tresponse %j', usage, error, response);
-    edebug(message);
+    const message = util.format('Failed reporting usage error for the' +
```
hsiliev: Shouldn't the message read something like `Failed reporting usage. Consecutive failures: %d`. We don't actually count the failures on each usage but the total consecutive.

PlamenDoychev: Fixed

PR #452, opened 2016-11-17
Temporarily disable bloom filter.

KRuelY:
#453


lib/utils/dataflow/src/index.js
```diff
@@ -169,7 +169,8 @@ const detectDup = function *(oid, ddup, odb) {
   debug('Checking for duplicate output doc %s', oid);
 
   const ddoc = ddup.has(oid);
-  if(ddoc === undefined) {
+  // Always check db because of the bloom filter bug.
```
hsiliev: We should point to #453

PR #461, opened 2016-11-21
Extend line length with 50% to 120 colums.

PlamenDoychev:
Dear colleagues, I propose to increase the number of columns per line with 50% to max 120. This will lead to:
1. Better readability of the code, as there are no strictly defined rules how to transfer text no new line.
2. 80 columns per line is very small for modern monitors. The majority of them is empty, and at the same time the code is with poor readability.
3. This will also reduce the number of lines in a single file.

### DIFF
--- a/tools/eslint/.eslintrc
+++ b/tools/eslint/.eslintrc
@@ -55,7 +55,7 @@ rules:
   no-eval: 2
   #require-yield: 2
   no-var: 2
-  max-len: [2, 80]
+  max-len: [2, 120]
   complexity: [2, 6]
   arrow-parens: [2, "always"]
 


KRuelY:
I don’t really mind the change. I’m not sure what others think about it though. Correct me if I’m wrong, but the reasons we are using the 80 lines are because the popular projects are using the rule (crockford, node, npm, airbnb, react) and so that you can display and compare multiple files side by side. Would be great if someone else can give their view!

hsiliev:
We need to reformat all source files to avoid situations where most of a file is 80 columns and we have line of two with 120.

So far we have not found a proper tool for the task.

abh1kg:
Also, interested in this purely because UI source code tends to get a little clumsy and having a constraint of 80 columns is really strict. Of course, if a preprocessing tool can help with reformatting it to a length of 80, it would be really nice

maximilien:
I don't mind this change but there needs to be a tooling proposed to do so and some human reviewer to do a pass making sure the tool did not result in some files having odd formatting.

Also, we'd needless to say need to verify that the resulting code works.

I'd be happy to volunteer doing this. But I need the following:

1. All committers to comment here that they agree with this. If not I will not proceed.
2. Some tooling suggestions to do so.
3. Another volunteer that will do a quick second pass (after mine) before we commit any change

So please comment here, otherwise I will ignore this and close in a week or so.

hsiliev:
+1 for the change

@astefanova made some research on how to reformat the code. Can you please shed some light on the topic?

astefanova:
Hi,

Me and @stoyanr researched a bit on that topic. We wanted to be able to format the code "on save", according to all eslint rules. 
Unfortunately we didn't find an appropriate solution.

We tried a few of the "automatic javascript code formatters" and neither of them seemed to be good enough.

The tools that were closest to our needs were:
* Eslint
We tried with the eslint inline tool `--fix`, but it cannot be used to fix all issues.

* [js-beautify](https://www.npmjs.com/package/js-beautify)
Here we found out that the most common-sensed decision in order to use this tool for automatic formatting would be to reformat all our code according to the js-beautify rules and afterwords rework a bit our eslint rules in order to comply.

maximilien:
+1

Can we merge this @astefanova?

astefanova:
@maximilien Did we choose an appropriate way/tool to refactor all the code?

hsiliev:
I would propose to use js-beautify and manually fix the remaining eslint issues. This would require some effort though.

maximilien:
+1 for js-beautify. Not sure the effort is worth it, however. I leave it to you all to debate this.

hsiliev:
https://github.com/prettier/prettier can get us closer to the desired state as it supports column width reformatting & integration with eslint at the same time

With it I was able to get a 0 errors (no manual interaction) for `dataflow` and `couchclient`

maximilien:
I also like the fact that `https://github.com/prettier/prettier` has significant following and activities.

hsiliev:
Closing in favor of #814

PR #463, opened 2016-11-24
Add prefix support

yulianedyalkova:
Suffix does not work with big installations (more that one instance per app), so we change it to prefix.


hsiliev:
This should help to support blue-green deploy and in general multiple deployments of Abacus.

PR #471, opened 2016-12-05
Dataflow reducer takes care of reducing data

rajkiranrbala:
The accumulator and aggregator apps should provide only the functions the produces outputs when the inputs are given. However, they took care of reducing a batch of input documents. Dataflow module is the place where the reduction should happen.

### DIFF
--- a/lib/metering/collector/src/index.js
+++ b/lib/metering/collector/src/index.js
@@ -373,6 +373,7 @@ const runCLI = () => {
 
 // Export our public functions
 module.exports = collector;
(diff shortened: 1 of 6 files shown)


lib/metering/collector/src/index.js
```diff
@@ -373,6 +373,7 @@ const runCLI = () => {
 
 // Export our public functions
 module.exports = collector;
+module.exports.noramalize = normalizeUsage;
```
KRuelY: normalize*

PR #475, opened 2016-12-07
Update deploy directory on build.

jmcarp:
It looks like https://github.com/cloudfoundry-incubator/cf-abacus/pull/472 changed the directory structure of the landscape directory, moving deploy config from landscape/abacus-config to landscape/abacus-config/deploy. This patch updates the concourse build script to copy the new deploy directory to the task output so that the correct manifests are used during deploy.

### DIFF
--- a/etc/concourse/scripts/cf-deploy-build
+++ b/etc/concourse/scripts/cf-deploy-build
@@ -15,7 +15,7 @@ if [[ $CONFIGURE = true ]]; then
 fi
 
 echo "Copying custom settings ..."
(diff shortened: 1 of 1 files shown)


hsiliev:
Hey @jmcarp,

Thanks for the fix. Can you also extract `abacus_config_dir` https://github.com/cloudfoundry-incubator/cf-abacus/pull/475/files#diff-c2a0c807f43b4c72eda54c6ddd375afcL13 outside configure block?

PR #478, opened 2016-12-09
Add AUTH_SERVER env variable to the manifest

shoudusse:
AUTH_SERVER environment variable is also needed for non secured
Abacus installation while requesting a token to Cloud Foundry


hsiliev:
Thanks for fixing the readme

PR #482, opened 2016-12-12
Improvements in concourse deployment scripts

georgi-lozev:
- create or update UAA clients
- db service instance deletion

Signed-off-by: Hristo Dobtchev <hristo.dobtchev@sap.com>

### DIFF
--- a/etc/concourse/scripts/cf-deploy-deploy
+++ b/etc/concourse/scripts/cf-deploy-deploy
@@ -2,6 +2,8 @@
 
 set -e -x
 
+SCRIPT_DIR="${BASH_SOURCE%/*}"
+
 echo "Environment:"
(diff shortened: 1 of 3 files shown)


sharms:
What do you think about making the creation of UAA clients optional?  We could submit a PR if needed, but were considering that Abacus should not have UAA admin permissions during it's lifecycle

hsiliev:
@sharms That would work provided you also disable cf-bridge in apps.rc. The reason we use `admin.read-only` is that we need to read app-usage events in the bridge.

As a minimum we need the abacus system client to support the pipeline processing.

hsiliev:
@sharms Any PR that improves the pipeline (UAA creation) is welcome.

PR #484, opened 2016-12-13
Allow operators to configure slack interval in concourse.

jmcarp:
So that `$SLACK` can be interpolated into cf manifests in `cf-deploy-build`.

### DIFF
--- a/etc/concourse/deploy-pipeline-vars.yml
+++ b/etc/concourse/deploy-pipeline-vars.yml
@@ -121,6 +121,9 @@ jwtalgo: algo
 # Use templates true/false
 abacus-configure: true
 
+# Slack interval
+slack: 10m
+
 # Abacus repo URI (or landscape project URI)
 landscape-git-repo: https://github.com/cloudfoundry-incubator/cf-abacus.git
 
--- a/etc/concourse/deploy-pipeline-vars.yml
+++ b/etc/concourse/deploy-pipeline-vars.yml
@@ -122,7 +122,7 @@ jwtalgo: algo
 abacus-configure: true
 
 # Slack interval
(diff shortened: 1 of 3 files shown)


etc/concourse/deploy-pipeline-vars.yml
```diff
@@ -121,6 +121,9 @@ jwtalgo: algo
 # Use templates true/false
 abacus-configure: true
 
+# Slack interval
+slack: 10m
```
hsiliev: I think we should have bigger slack interval. The slack interval control several max delays:
* max downtime time for Abacus
* max delay of any resource provider
   * max time of app usage events for cf-bridge
   * max polling interval for some providers

Having the default to 10 minutes might be too restrictive for some use-cases. We usually set this to 3 or even 5 days, which is the time needed to fix a tough productive instance problem after a weekend, no matter if there is 24x7 or not :)

jmcarp: Makes sense, I changed the default to "5D" to be safe. I used "10m" initially because it seems to be the default when `SLACK` is unset: https://github.com/cloudfoundry-incubator/cf-abacus/blob/a9c064745fe05e1784c0e0f4083f5ff9edb3c116/lib/aggregation/accumulator/src/index.js#L96-L99. Should that be changed too?

hsiliev: I think that the default is the minimal sane value. It makes the testing easier and keeps the data as close to the month boundaries as possible.
The concrete productive slack interval varies depending on your resource providers behavior

I personally would like to keep the default as it is to avoid complicating the test setup.

PR #486, opened 2016-12-15
Add optional pre-build step in deploy pipeline

astefanova:

### DIFF
--- a/etc/concourse/scripts/cf-deploy-build
+++ b/etc/concourse/scripts/cf-deploy-build
@@ -10,11 +10,14 @@ cp -r landscape/cf-abacus/. built-project
 
 abacus_config_dir="landscape/abacus-config/deploy"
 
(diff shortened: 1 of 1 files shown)


etc/concourse/scripts/cf-deploy-build
```diff
@@ -10,6 +10,10 @@ cp -r landscape/cf-abacus/. built-project
 
 abacus_config_dir="landscape/abacus-config/deploy"
 
+if [[ -e bin/pre-build ]]; then
+  ./bin/pre-build $abacus_config_dir
```
hsiliev: I think this should be `pre-configure` and executed only if we want to replace templates

hsiliev:
Since we anyway copy the configuration to Abacus it actually make sense to have the pre-configure script executed even without replacing templates.

We should consider renaming the `CONFIGURE` flag

PR #492, opened 2017-01-02
Add prefix to test pipeline

astefanova:


etc/concourse/scripts/cf-test-test
```diff
@@ -1,14 +1,80 @@
 #!/bin/bash
 
-set -e
+set -e -x
+
+APPS=()
```
hsiliev: We should extract `getApps` and `restartApps` to enable their reuse across scripts

etc/concourse/scripts/cf-test-test
```diff
@@ -1,14 +1,80 @@
...
-  echo "Starting CF test ..."
-  npm run cftest
-popd
\ No newline at end of file
+  echo "Deleting all apps in $CF_ORG/$CF_SPACE ..."
+  getApps
+  echo ${APPS[@]} | xargs -n1 | xargs -P 5 -i cf delete -f -r {}
+
+  echo "Deploying Abacus ..."
+  npm run cfstage default
```
hsiliev: `cftest` uses `medium` profile. Do we need to change the profile to `default`

PR #516, opened 2017-01-20
Add admin user to cf-deploy-start job

nmaslarski:
Signed-off-by: Nikolay Maslarski <nikolay.maslarski@sap.com>


hsiliev:
The admin might be needed by hooks, executed from the pipeline.
For example we might want to restrict the access to Abacus apps (except collector and reporting).

Issue #228, opened 2016-02-08
Enhance abacus to enable custom provider of current time/date


piotrprzybylski:
Could you provide ability to supply a date/time provider which could overwrite getting the current date/time by Abacus. This would enable automated integration testing validating long time running usage metering without the need to modify system clock. We are looking effectively at the functionality offered by sinon timers for integration testing.

jsdelfino:
Any volunteer to implement this?

betafood:
I'm wondering if we may want to use a library like [moment.js](https://github.com/moment/moment) for our general dating purposes (since it may simplify some of the time window calculation logic).

For this particular scenario, the time source can easily be [customized](http://momentjs.com/docs/#/customization/now/). This would allow something like an environment variable to easily decide the "current" time.

Although, switching to use this would most likely be a pretty involved change.

Thoughts?

hsiliev:
Would a simple time encoded in an environment variable do here? If I got
this right the tests should be able to pass a custom function. We might
need a bit more complex solution (module name?), but it seems doable.

2016-04-13 21:43 GMT+03:00 Benjamin Cheng <notifications@github.com>:

jsdelfino:
Sounds like there consensus on using an env variable (wether we use moment.js or not, which could also be covered in a different issue). Do you guys want that env variable to configure a fixed time (so the time actually does not pass), or do you want the the env variable to be an offset (whereby the time continues to pass normally but is offset to 3 months back to match the times of your test data for example)?

jsdelfino:
@piotrprzybylski, you initially opened this issue. Any thoughts on my last comment?

hsiliev:
We want to introduce:
* `abacus-moment` wrapper that handles the custom-time-provider and adds an offset from environment variable
* eslint plugin to check for the use of `Date`

Then we will replace all uses of Date with the new `abacus-moment`

We have a working version that has 6-8% performance penalty due to the replacement of native-backed `Date` with JS-based `moment`

PR #520, opened 2017-01-24
Use abacus-moment in util modules

hsiliev:
#228 
[#113294537]


KRuelY:
Hi, thanks for the pull requests!

The only one I saw missing from moving from using Date to moment is timewindow's test. Aside from that I think you pretty much covered everything.

PR #531, opened 2017-02-07
Remove container plan

nmaslarski:
Signed-off-by: Nikolay Maslarski <nikolay.maslarski@sap.com>

It seems to be identical to linux-rating-plan.


KRuelY:
Yeah, this is a duplicate that was probably used for testing. Thanks for catching it.

PR #537, opened 2017-02-13
Remove path param plan_id from create endpoint in Provisioning

astefanova:


KRuelY:
Looks good!

PR #542, opened 2017-02-16
Fix reporting authorization

### LINKED ISSUE #535: Organization report breaks usage isolation
Resource providers can see other resource providers date through the organization report.

We can limit the org report access and require system token access.
The resource instance report should be accessible only to the respective resource providers that have token with `.write` scope.

nmaslarski:
Fixes #535
[delivers #139495947]

### DIFF
--- a/test/perf/src/test/test.js
+++ b/test/perf/src/test/test.js
@@ -337,12 +337,12 @@ describe('abacus-perf-test', () => {
       brequest.post('http://localhost:9080/v1/metering/collected/usage',
         extend({}, opt(objectStorageToken),
           { body: usageTemplate(o, ri, i) }), (err, val) => {
-          expect(err).to.equal(undefined);
-          expect(val.statusCode).to.equal(201);
-          debug('Completed submission org%d instance%d usage%d',
-            o + 1, ri + 1, i + 1);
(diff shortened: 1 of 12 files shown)


test/perf/src/test/test.js
```diff
@@ -293,16 +293,43 @@ describe('abacus-perf-test', () => {
       ]
     };
 
+    const token2 = {
```
hsiliev: Can we rename the token to "systemReadToken" and the one above to something like "objectStorageWriteToken" (or shorter) to make the purpose of the tokens more obvious ?

test/perf/src/test/test.js
```diff
@@ -293,16 +293,43 @@ describe('abacus-perf-test', () => {
...
+    };
+
     // OAuth bearer access token signed using JWTKEY and
     // default algorithm (HS256)
     const auth = process.env.SECURED === 'true' ?
       jwt.sign(token, process.env.JWTKEY, {
         algorithm: process.env.JWTALGO,
         expiresIn: 43200
       }) : undefined;
+    const auth2 = process.env.SECURED === 'true' ?
```
hsiliev: Same here - we need better names for the variables.

PR #543, opened 2017-02-18
Execute each test file in a separate process (#525)

### LINKED ISSUE #525: abacus-mocha tests are executed in common process
`abacus-mocha` executes all `*test.js` files under the same process. This leads to two problems:
* test output is not produced until all tests complete
* tests share the same module cache

If we run the tests in separate processes this will simplify the test logic and give us better dev-turnaround time.

stoyanr:
Fixes #525 
[delivers #138810633]


tools/mocha/src/index.js
```diff
@@ -12,130 +12,22 @@ if(process.env.LONGJOHN)
...
+var async = require('async');
 
-var map = _.map;
-var values = _.values;
 var contains = _.contains;
 var memoize = _.memoize;
+var extend = _.extend;
 
-/* eslint no-process-exit: 1 */
+/* eslint no-process-exit: 0 */
```
stoyanr: No, this actually disables the check, which is by default on. If we remove this line, eslint will start complaining.

PR #544, opened 2017-02-19
Replace import / export with require and remove babel (#538)

### LINKED ISSUE #538: Babel not really needed with node 7
Using babel to transpile code has some disadvantages:
* Slower build process and development cycle
* Harder tracking of error messages since line numbers point to the transpiled code
* Interactive debugging difficult or impossible

It would be better if we could get rid of babel and run our source code directly. With node 7, this is already possible. The only not supported feature is import / export. These are used in abacus in only the following 4 modules: `iterable`, `learning`, `matrix`, and `rewire`. Getting rid of them is pretty easy (I did it today, took me a couple of hours).

I did the following small PoC on my laptop:
* Disabled the babel es2015 preset (babel is still running, but not transpiling)
* Replaced all import / export statements with require / module.exports
* Ran all unit tests successfully
* Ran all integration tests successfully

stoyanr:
Fixes #538
[delivers #139710685]


hsiliev:
Any idea why the coverage dropped 36%? Could this indicate a problem with the way we collect it after removing babel?

stoyanr:
Yes, there was some code in the coverage tool that relied on lib instead of src. Should be fixed now.

PR #554, opened 2017-02-28
Make replay run in parallel with app.listen

### LINKED ISSUE #552: Replay logic cannot run more than 1 minute
Replay logic on Diego is restricted to 1 minute since:

1. the default health check is port with timeout of 1 minute
2. Abacus apps start to replay 
3. only after replay finishes we open the port

We can do one of the following:
* change the health check to `http` or `process`
* start listening in parallel with the replay
* extract the replay as Diego task and delegate the responsibility to to external party (blue-green deploy for instance)

nmaslarski:
Fix #552

### DIFF
--- a/lib/aggregation/accumulator/src/index.js
+++ b/lib/aggregation/accumulator/src/index.js
@@ -352,10 +352,12 @@ const runCLI = () => {
   // Create the app, replay any failed events, and listen
   const app = accumulator();
   app.listen();
-  debug('Replaying for accumulator');
-  dataflow.replay(app.reducer, 0, () => {
-    debug('Replay finished successfully for accumulator');
-  });
+  if (cluster.isWorker()) {
+    debug('Replaying for accumulator');
+    dataflow.replay(app.reducer, 0, () => {
+      debug('Replay finished successfully for accumulator');
+    });
+  }
 };
 
 // Export our public functions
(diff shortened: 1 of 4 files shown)


lib/aggregation/accumulator/src/index.js
```diff
@@ -351,9 +351,13 @@ const runCLI = () => {
 
   // Create the app, replay any failed events, and listen
   const app = accumulator();
-  dataflow.replay(app.reducer, 0, () => {
-    app.listen();
-  });
+  app.listen();
+  if (cluster.isWorker()) {
```
hsiliev: We can remove this check since:
1. As @KRuelY noticed, we already have a check in dataflow that prevents `replay` running more than once
2. We should be running `replay` on master since workers can be > 1

lib/aggregation/accumulator/src/index.js
```diff
@@ -351,9 +351,13 @@ const runCLI = () => {
...
+    debug('Replaying for accumulator');
```
hsiliev: Debug already logs the abacus-accumulator location, so we might simply log something like `Starting replay...`

lib/aggregation/accumulator/src/index.js
```diff
@@ -351,9 +351,13 @@ const runCLI = () => {
...
+    dataflow.replay(app.reducer, 0, () => {
+      debug('Replay finished successfully for accumulator');
```
hsiliev: We need to check for error. Maybe something like:
```
dataflow.replay(app.reducer, 0, (err) => {
  debug('Replay completed %s', err ? 'with error' : 'successfully');
});
```	

The error and the docs are already logged by dataflow.

PR #556, opened 2017-02-28
Bump node version to 6.10.0

hsiliev:


hsiliev:
Change will be part of babel removal

PR #576, opened 2017-03-14
Fix mongo and couch connection pooling issues

stoyanr:
Connection pooling in both couch and mongo clients was broken when multiple parallel operations (e.g. during a batch op) attempted to get the same connection from an initially empty pool. Instead of creating just one connection per partition and then caching and reusing it, as many connections were created as there were parallel operations started, even for the same partition. This resulted in crashing mongo during the execution of unit tests in some cases due to reaching the max number of connections.

This issue is now fixed by adding locking to ensure that the initial creation of each connection happens only once. In addition, the code in both clients is refactored to remove code duplication in the `pool` function, to comply with existing naming conventions, and to ensure that the similarity between both clients is preserved as much as possible. For mongo, pooling by read / write mode is removed as it doesn't seem to be needed.

### DIFF
--- a/lib/utils/couchclient/src/index.js
+++ b/lib/utils/couchclient/src/index.js
@@ -10,6 +10,7 @@ const memdown = require('memdown');
 const moment = require('abacus-moment');
 const perf = require('abacus-perf');
 const url = require('url');
+const lock = require('abacus-lock');
 
 // Configure needed PouchDB functionality
 const PouchDB = require('pouchdb');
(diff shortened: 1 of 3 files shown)


lib/utils/couchclient/src/index.js
```diff
@@ -435,63 +436,54 @@ const couchclient = (part, uri, cons) => {
...
+    // Get a pooled db partition handle for a single partition
+    const pooldb = (p, rw, cb) => {
+      const u = dbopt.uri(p);
+      debug('Using db %s in %s mode', puri(u), rw);
+
+      // Return a memoized db partition handle or get and memoize a new one
+      // from the given db constructor. DB handles are keyed by db uri and
+      // read/write operating mode
+      const dbkey = [u, rw].join('-');
+      lock(dbkey, (err, unlock) => {
```
hsiliev: Do we need to handle error from lock here?

hsiliev: How do we ensure this does not regress? Do we have a new test (or adapted one) to cover this case?

stoyanr: The lock module will always pass `undefined` as error. To ensure that this doesn't regress, I will amend the PR to increase the number of documents used by the corresponding unit test from 100 to 1000. Without proper pooling, this should fail.

PR #582, opened 2017-03-17
Make concourse scripts configurable

yulianedyalkova:
Add ABACUS_PREFIX property.


hsiliev:
Can we add a line or two in the concourse pipeline readme for the purpose of the PREFIX?

PR #589, opened 2017-03-24
Extend db clients with drop partitions capabilities

stoyanr:
Signed-off-by: Plamen Doychev <plamen.doychev@sap.com>

### DIFF
--- a/lib/utils/couchclient/src/index.js
+++ b/lib/utils/couchclient/src/index.js
@@ -107,6 +107,45 @@ const key = (uri) => {
   return undefined;
 };
 
+// List all databases on the specified server
+const listDatabases = (server, cb) => {
+  debug('Listing databases on %s', server);
+  request.get(':server/_all_dbs', {
+    server: server
+  }, (err, val) => {
+    if(err) {
+      edebug('Failed to list databases on %s because of %o', server, err);
+      return cb(err);
+    }
+
+    return cb(null, val.body ? val.body : []);
+  });
+};
+
+// List all in-memory databases on the specified server
+const listInMemoryDatabases = (server, cb) => {
+  debug('Listing in-memory databases on %s', server);
+  const globalStoreLocations = Object.keys(memdown('_$nonexistent$_')._store);
+  const prefix = '$' + (server ? server + '/' : '');
+  return map(globalStoreLocations, (key) => key.replace(prefix, ''));
+};
+
+// Drop the specified database on the specified server
+const dropDatabase = (server, database, options, cb) => {
+  debug('Dropping database %s on %s', database, server);
+  const dbName = server ? [server, database].join('/') : database;
+  const db = new PouchDB(dbName, options);
+  db.destroy((err) => {
+    if (err) {
+      edebug('Failed to drop database %s on %s because of %o',
+        database, server, err);
+      return cb(err);
+    }
+
+    return cb(null);
+  });
+};
+
 // Return a db uri naming function configured with a db uri name prefix
 const dburi = (server, name) => {
   if(!server)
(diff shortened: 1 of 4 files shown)


etc/concourse/scripts/cf-deploy-stage
```diff
@@ -34,4 +34,5 @@ if [ "$BIND_DB_SERVICE" == "true" ]; then
   bindService ${ABACUS_PREFIX}abacus-usage-reporting $DB_AGGREGATOR,$DB_ACCUMULATOR
   bindService ${ABACUS_PREFIX}abacus-cf-bridge,${ABACUS_PREFIX}abacus-cf-renewer $DB_BRIDGE
   bindService ${ABACUS_PREFIX}abacus-provisioning-plugin,${ABACUS_PREFIX}abacus-account-plugin $DB_PLUGINS
+  bindService ${ABACUS_PREFIX}abacus-housekeeper $DB_COLLECTOR,$DB_METER,$DB_ACCUMULATOR,$DB_AGGREGATOR,$DB_BRIDGE,$DB_PLUGINS
```
hsiliev: Hey shouldn't this be a part of another PR that introduces the housekeeper app?

lib/utils/couchclient/src/index.js
```diff
@@ -699,65 +738,84 @@ const couchclient = (part, uri, cons) => {
...
+  };
+
+  if(/:/.test(server))
+    dropServerDatabases(server, regex, {}, localOnly, lcb);
+  else
+    dropInMemoryDatabases(server, regex, inMemoryOpt, lcb);
+};
+
+// Delete all partitions that match the specified regex
+const deletePartitions = (server, regex, cb) => {
```
hsiliev: Not sure if partition has any meaning in Couch. Perhaps we can leave only `drop` for couch?

lib/utils/couchclient/src/index.js
```diff
@@ -699,65 +738,84 @@ const couchclient = (part, uri, cons) => {
...
+  dropDatabases(server, regex, false, cb);
```
hsiliev: Running demo or smoke test vs productive instance will result in dropped database. We should disable this code-line by default and enable it only on request (via env.variable for instance) to protect productive data from accidental deletion.

lib/utils/couchclient/src/index.js
```diff
@@ -107,6 +107,45 @@ const key = (uri) => {
   return undefined;
 };
 
+// List all databases on the specified server
```
stoyanr: It's there because this is the order in the mongoclient. And in the mongoclient, the order is such because these methods are also used elsewhere (e.g. findServer).

etc/concourse/scripts/cf-deploy-stage
stoyanr: OK, I will move that to a different PR.

lib/utils/couchclient/src/index.js
stoyanr: "Partition" is a unit that's used for partitioning. With Couch, partitions are always databases. With Mongo, however, partitions can be databases or collections, and `deletePartitions` will correctly determine what to delete in both cases. Therefore, the naming "partition" is important and since both clients should have the same API, both should have such a method.

The method `deletePartitions` is intentionally different from `drop` because:
* It always works with partitions, which might be different from databases (with Mongo)
* It deletes partitions also on remote servers, while `drop` deletes databases on local servers only.

lib/utils/couchclient/src/index.js
stoyanr: No. This method is not used (and should not be used) anywhere except the new `housekeeper` module which will be part of `cf-abacus-broker`.

I will add more comments in the code to warn about the implications of using this method. In the future, we can also add an `eslint` check to make it even harder to misuse it accidentally.

PR #609, opened 2017-04-21
Move healthcheck endpoint from eureka to webapp module

### LINKED ISSUE #601: Replace OPTIONS endpoint with GET /health
OPTIONS endpoint in webapp (all Abacus apps basically) exposes the following headers:
```
X-App-Index →0
X-App-Name →abacus-usage-collector
X-App-Version →0.0.6-dev.8
X-Backside-Transport →FAIL FAIL
X-Heap-Used →229022408
X-Instance-Id →
X-Instance-Index →0
X-Node-Version →v6.10.0
X-Powered-By →Express
X-Process-Id →75
X-Response-Time →60.748
X-Uptime →331072.852
```

There are several issues with this approach:
* Node and app versions can be used for CVE lookup
* memory can be used for DoS attacks
* OPTIONS is not supported by a lot of tools

We should replace the OPTIONS request with `GET /health` that returns 200 if the app is healthy and 500 if not. 

The new endpoint can return the same info as above, protected with (optional) basic authentication. We can reuse the `HEALTHCHECK_SECURED` variable used in eureka healthcheck.

martin-d-aleksandrov:
Fixes #601
[delivers #143531977]

### DIFF
--- a/lib/utils/express/src/test/test.js
+++ b/lib/utils/express/src/test/test.js
@@ -11,6 +11,8 @@ const express = require('..');
 /* jshint undef: false */
 /* jshint unused: false */
 
+debug.enable('abacus-express');
+
 describe('abacus-express', () => {
   let clock;
   let exit;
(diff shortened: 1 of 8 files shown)


lib/utils/express/src/test/test.js
```diff
@@ -11,6 +11,8 @@ const express = require('..');
 /* jshint undef: false */
 /* jshint unused: false */
 
+debug.enable('abacus-express');
```
hsiliev: As far as I can see we enable debugging for just a single test. Can we have it wrapped in context and set/unset with before/after?

PR #611, opened 2017-04-21
Fail when creating or updating a service failed

MarcBeckerSAP:
In createOrUpdateService error checks where completely turned off. However this should only be the case for the check of the existance of the service via the "cf service" command. Otherwise errors during "cf create-service" or "cf update-service" go unnoticed. 
Even when these commands trigger an asynchronous action the return code is 0, which allows error checks to be enabled during the execution of these commands.

### DIFF
--- a/etc/concourse/scripts/common-functions
+++ b/etc/concourse/scripts/common-functions
@@ -60,16 +60,18 @@ function waitForServices {
 }
 
 function createOrUpdateService {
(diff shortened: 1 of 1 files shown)


MarcBeckerSAP:
Please close and reopen PR. Missed to publish my corporate CLA membership, which is done now.

PR #612, opened 2017-04-24
Add distinct values query in DB clients

nmaslarski:


lib/utils/couchclient/src/index.js
```diff
@@ -400,6 +400,10 @@ const couchclient = (partitionFn, uriFn, connectionFn, errorDBFn) => {
           });
     },
 
+    distinctValues: () => {
+      throw Error('Not Implemented');
```
hsiliev: It is better to call back with error, than throw here

lib/utils/couchclient/src/test/test.js
```diff
@@ -69,6 +69,11 @@ describe('abacus-couchclient', (d) => {
     };
   };
 
+  it('should throw error when call distinctValues', () => {
```
hsiliev: "distinctValues errors" or "distinctValues calls back with error" ?

lib/utils/couchclient/src/index.js
nmaslarski: The idea is that this way it will be more disruptive and the execution will be aborted. The call back error is easier to be neglected.

PR #623, opened 2017-05-02
Concourse cf-deploy-infra script should not wait for external services

yulianedyalkova:
Currently if there are other services deployed in the same org and space, the cf-deploy-infra script will fail. This PR proposes a fix that will make the script wait only for the abacus services.

### DIFF
--- a/etc/concourse/scripts/common-functions
+++ b/etc/concourse/scripts/common-functions
@@ -48,15 +48,19 @@ function restartAppsWithRetry {
 
 function waitForServices {
   echo "Waiting for DB creation or update ..."
-  until [ "$(cf services | grep '\(create\|update\) succeeded\|\(create\|update\) failed' | wc -l)" == "$1" ]; do
-    sleep 3s
+  services=$1
+  for service in "${services[@]}"
+  do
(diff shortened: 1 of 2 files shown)


etc/concourse/scripts/common-functions
```diff
@@ -48,23 +48,27 @@ function restartAppsWithRetry {
...
-  until [ "$(cf services | grep '\(create\|update\) succeeded\|\(create\|update\) failed' | wc -l)" == "$1" ]; do
+  services=$1
+  for service in "${services[@]}"
+  do
+    while [[ -z $(cf service $service | grep '\(create\|update\) succeeded\|\(create\|update\) failed') ]]; do
     sleep 3s
   done
-  if [ "$(cf services | grep '\(create\|update\) succeeded' | wc -l)" == "$1" ]; then
-    echo "DB creation or update finished successfully."
+  if [[ -n $(cf service $service | grep '\(create\|update\) succeeded') ]]; then
```
hsiliev: We need to indent the if so it is more visible the checks are executed in the for. Not sure if moving the do (line 53) next to the for cycle would help to achieve this.

PR #640, opened 2017-05-18
Delete obsolete apps during deployment

stoyanr:


etc/concourse/scripts/common-functions
```diff
@@ -121,3 +121,12 @@ function mapRoutes {
     done
   fi
 }
+
+function deleteObsoleteApps {
+  local APP_NAME=$1
+  local MAX_INSTANCES=$(expr $2 + 4)
```
hsiliev: What's the logic to have `+ 4`?

Issue #293, opened 2016-04-11
Abacus releases should be reproducible


hsiliev:
Currently we only tag the releases, without fixing the dependencies. Therefore it is not easy to checkout old version (say 0.0.3) and build it. Sometimes this is not even possible due to the open ranges used for module dependencies.

We should publish the exact versions that we used to validate and release Abacus for each released version.

I believe we can do this with `npm shrinkwrap` to fix the dependencies, then store the `npm-shrinkwrap.json` in a `release` (or the `etc`) directory for future use. This file will be versioned by git, so checking out a specific release will also download the exact set of dependencies.

jsdelfino:
I was planning to do this for the next release (which I'd like to cut next week during the CF summit) but I'm not sure we'll be able to use npm shrinkwrap yet for that release, as I believe folks were running into errors running npm shrinkwrap. I'll try and see if these errors are still there.

PR #652, opened 2017-06-02
Fix dependencies with shrinkwrap and introduce deps update script

HristoDobchev:
This change fixes the dependencies versions so that we can have more predictable behavior when we release and provision productive instances.
The change also introduces an "update" npm script that regenerates the npm-shrinkwrap.json thus updating all dependencies.

Related to issue #293


hsiliev:
Will this handle all apps that we cfpack & cfpush? We might need to do the same for all modules

package.json
```diff
@@ -38,6 +38,7 @@
     "build": "npm run provision && npm test",
     "clean": "git clean -df",
     "rebuild": "npm run clean && npm run build",
+    "update": "rm -rf node_modules && rm npm-shrinkwrap.json && npm run provision && npm shrinkwrap && bin/cfupdate",
```
hsiliev: Can we extract this line as a script in bin? Perhaps merge it with cfupdate one and rename everything to module-update?

bin/cfupdate
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
hsiliev: Might be better to have this as `module-update`. Now it implies that we only do this when working with cf which is not 100% true.

Issue #617, opened 2017-04-27
Wrong abacus reporting in case of wrong/missing organization id.


PlamenDoychev:
Abacus reporting return status code 200 and empty JSON in case of wrong or missing organization guid is passed (see below). 
According to the documentation this status code should be returned in case of success and JSON object should contain usage data. In case of usage is not found for organization, 404 should be returned by specification again.

No report data! Getting original response:
HTTP/1.1 200 OK
Content-Length: 192
Content-Type: application/json; charset=utf-8
Date: Thu, 27 Apr 2017 12:23:23 GMT
Etag: W/"c0-qrzzfAcm3kK+9e40C6D51wa3KWg"
Vary: Accept-Encoding
X-App-Index: 5
X-App-Name: abacus-usage-reporting
X-App-Version: 0.0.6-dev.8
X-Heap-Used: 36078256
X-Instance-Id: 38e9cbe3-1120-4bec-6ba4-0b075496c849
X-Instance-Index: 0
X-Node-Version: v7.8.0
X-Powered-By: Express
X-Process-Id: 55
X-Response-Time: 62.972ms
X-Uptime: 104005.032
X-Vcap-Request-Id: 394107c8-b598-47c8-61b3-6ec4d664e897

{"id":"k/b51c5fd3-0b4c-4dcb-8968-f010c5b60123/t/0001493295803198","organization_id":"b51c5fd3-0b4c-4dcb-8968-f010c5b60123","start":1491004800000,"
end":1493295803198,"resources":[],"spaces":[]}

Regards,
Plamen Doychev

martin-d-aleksandrov:
It seems there is no issue here. The reporting app returns 404 in case the organization does not exists. The reporting checks this by calling the account plugin, so it should implement the correct check.

PR #657, opened 2017-06-12
Fix error status code returned from reporting app

martin-d-aleksandrov:
https://github.com/cloudfoundry-incubator/cf-abacus/issues/617


hsiliev:
Can we name json resources with more meaningful names?

martin-d-aleksandrov:
@hsiliev These .json files where extracted out of the already existing test in the reporting. Unfortunately I could not extract info on the semantics of these objects, that's why I left the names unchanged

PR #664, opened 2017-06-21
Extend oauth to work with custom endpoint

astefanova:
1. Extend OAuth to work with custom endpoint and token path
2. Use "POST" method when making access token requests according to [RFC6749](https://tools.ietf.org/html/rfc6749#section-3.2)
3. Cache the token endpoint retrieved from CF API

Signed-off-by: Stoyan Rachev <stoyanr@gmail.com>

### DIFF
--- a/lib/utils/oauth/src/index.js
+++ b/lib/utils/oauth/src/index.js
@@ -4,6 +4,7 @@
 
 const _ = require('underscore');
 const jwt = require('jsonwebtoken');
+const lru = require('abacus-lrucache');
 const request = require('abacus-request');
 const retry = require('abacus-retry');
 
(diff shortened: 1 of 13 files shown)


lib/utils/oauth/src/index.js
```diff
@@ -262,32 +262,40 @@ const cache = (authServer, id, secret, scopes) => {
...
           return;
         }
 
-        // Get token from the OAuth token endpoint
-        newToken(endpoint, id, secret, scopes, (err, t) => {
+        debug('Got OAuth bearer access token');
+        token = t;
+        // Convert token expiration period from seconds to milliseconds
+        // Without conversion, the token will refresh every 2 minutes
+        refresh(Math.max(token.expires_in * 1000 - 120000, 120000));
```
hsiliev: Do we always have seconds in the response for custom endpoints?

astefanova: According to OAuth 2.0 standard:
`expires_in: The lifetime in seconds of the access token.`
[see here for more](https://tools.ietf.org/html/rfc6749#section-4.2.2)

PR #672, opened 2017-06-29
Free event loop during report generation

HristoDobchev:


lib/aggregation/reporting/src/index.js
```diff
@@ -130,6 +133,10 @@ const summarizefn = (metrics, metric) => {
   return filter(metrics, (m) => m.name === metric)[0].summarizefn;
 };
 
+const setTimeoutEx = yieldable((delay, cb) => {
+  setTimeout(cb, delay);
```
hsiliev: Can we do this as setImmediate?

PR #676, opened 2017-07-06
Add options to abacus-client

astefanova:


lib/utils/client/src/test/reporting-test.js
```diff
@@ -366,22 +364,33 @@ const tests = (secured) => {
...
             responseBody = generateResponse(409, {}, errorResponseBody);
 
             errors.consecutiveReportFailures = failures;
 
-            reporter.reportUsage(appUsage, token, (error, res) => {
-              expect(error).to.equal(undefined);
-              expect(res).to.equal(responseBody);
-              done();
-            });
+            if(isEmpty(options))
```
hsiliev: The branching logic in reporting about callback/options should handle this I think. If so do we need this check?

astefanova: This is done in order to do different checks for the two cases - 201 and 409 status codes.

PR #677, opened 2017-07-06
Update abacus resource provider guide

PlamenDoychev:


hsiliev:
Changes are now in Abacus wiki

PR #702, opened 2017-08-07
Fix exceeded throttle

svgradev:


lib/aggregation/reporting/src/test/reporting-test.js
```diff
@@ -151,81 +151,110 @@ describe('abacus-usage-report', () => {
...
+            undefined, undefined,
+          true), [planAReport, planBReport], [consumer1, consumer2]);
+      });
+
+      it('successfully when throttle limit is not exceeded', (done) => {
+
+        verify(false, () => verify(true, done));
+      });
+
+      // Test would fail (time-out) when retrieveUsage doesn't have a callback
```
hsiliev: We can remove this comment as this is the test used to verify the addition of the callback anyway

lib/aggregation/reporting/src/test/reporting-test.js
```diff
@@ -151,81 +151,110 @@ describe('abacus-usage-report', () => {
...
+      // as an argument and doesn't call that callback
+      context('when throttle limit is exceeded', () => {
+        before(() => {
+          process.env.THROTTLE = 10;
+          delete require.cache[require.resolve('..')];
+          report = require('..');
+        });
+
+        it('successfully', (done) => {
```
hsiliev: Should this be 'succeeds' to fit better with the context description?

hsiliev:
Mongo build fails due to change in Travis to Trusty
We need to adapt our build to install the proper packages to get this running

hsiliev:
https://github.com/cloudfoundry-incubator/cf-abacus/pull/703 tries to fix this build

PR #714, opened 2017-08-18
Adding couch-based connect session store

abh1kg:
This enables us to support non-sticky, serialized sessions for the Abacus service
dashboard, deployed with couchdb. This change includes:
- Addition of query method to couchclient to run map/reduce queries
- Addition of couch based sessionStore for session management

### DIFF
--- a/lib/utils/couchstore/src/index.js
+++ b/lib/utils/couchstore/src/index.js
@@ -1,0 +1,150 @@
+'use strict';
+/*  eslint-disable max-len,no-param-reassign*/
+const moment = require('abacus-moment');
+const partition = require('abacus-partition');
+const couchclient = require('abacus-dbclient');
+const debug = require('abacus-debug')('abacus-couchstore');
+const _ = require('underscore');
+
+const conectCouch = (connect) => {
+  let Store = connect.Store;
+
+  class CouchStore extends Store {
+
+    constructor(options) {
+      super();
+      let dbHandle = couchclient(partition.singleton, couchclient.dburi(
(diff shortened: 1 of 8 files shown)


lib/utils/couchstore/src/index.js
```diff
@@ -0,0 +1,150 @@
+'use strict';
+/*  eslint-disable max-len,no-param-reassign*/
```
hsiliev: Can you please adhere to the defined chars limit across the Abacus sub-modules?
We intend to increase this 80 chars to 120 soon, but this is yet another reason to leave the max-len out of disable list, so we can validate everything is ok.

lib/utils/couchstore/src/test/test.js
```diff
@@ -0,0 +1,142 @@
...
+const dbclient = require('abacus-dbclient');
+const moment = require('abacus-moment');
+const server = () => process.env.DB;
+
+describe('abacus-couchStore', () => {
+  let store;
+  let dbName = 'abacus-test-dashboard';
+  let cookie = { name: 'test-cookie', cookie: { maxAge: 5000 } };
+
+  before(() => {
```
hsiliev: You need to add a logic similar to this one: https://github.com/cloudfoundry-incubator/cf-abacus/blob/62578a4dc92030fe83ac6ae6249ccbb53cbc314d/lib/utils/couchclient/src/test/test.js#L39-L49

This will prevent the tests from running with `mongoclient` or any future clients.

PR #720, opened 2017-08-24
Implement cfpush retry mechanism

PlamenDoychev:


tools/mocha/package.json
```diff
@@ -38,7 +38,7 @@
     "istanbul": "^0.4.0",
     "longjohn": "^0.2.9",
     "mocha": "^3.2.0",
-    "sinon": "^1.14.1",
+    "sinon": "=2.4.1",
```
hsiliev: Shouldn't we have this as `^2.4.1`? Our update script should take care to fix this version

PlamenDoychev: Actially, no because the next version of sinon library is 3.0.0 and we hit some problems with it. That's why we want to stick exactly to the latest revision from 2 -nd version.

hsiliev: `^2.4.1` should be equal to `>=2.4.1 <3.0.0` according to https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004

If we somehow end-up with using 3.0.0 then I think we have another problem here.

georgi-lozev: +1 for ^2.4.1

PR #729, opened 2017-09-08
Change node version above 8

denicaM:
Change node.js version to 8.0.0.
(npm version is still <5.0.0)


.travis.yml
```diff
@@ -1,8 +1,12 @@
 language: node_js
-node_js: "6.10.0"
+node_js: "8.0.0"
```
hsiliev: We should change that to the latest 8.4.0?

demo/client/package.json
```diff
@@ -41,7 +41,7 @@
     "abacus-publish": "file:../../tools/publish"
   },
   "engines": {
-    "node": ">=6.10.0",
+    "node": ">=8.0.0",
```
hsiliev: We should have these as `>=8.4.0` as well to get the latest fixes

hsiliev:
@denicaM Can you please publish your SAP org membership so we can make dreddbot happy?

PR #740, opened 2017-09-20
Increase CC poll period and page size

HristoDobchev:
- reduce load on CF cloud controller by increasing the period at which its polled for new events and at the same time increase the page size (also parametrize it via env variable)

### DIFF
--- a/lib/cf/applications/src/index.js
+++ b/lib/cf/applications/src/index.js
@@ -59,7 +59,7 @@ const guidMinAge = parseInt(process.env.GUID_MIN_AGE) || 60000;
 const orgsToReport = process.env.ORGS_TO_REPORT ?
   JSON.parse(process.env.ORGS_TO_REPORT) : undefined;
 const lastRecordedGuid = process.env.LAST_RECORDED_GUID;
-const resultsPerPage = parseInt(process.env.PAGE_SIZE) || 100;
+const resultsPerPage = parseInt(process.env.API_PAGE_SIZE) || 100;
 
 const reportingConfig = {
   minInterval     : minIntervalTime,
--- a/lib/cf/applications/src/index.js
+++ b/lib/cf/applications/src/index.js
@@ -53,12 +53,13 @@ const db = throttle(retry(breaker(batch(
     'abacus-cf-bridge'))
 ))));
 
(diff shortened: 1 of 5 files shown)


lib/cf/applications/src/index.js
```diff
@@ -53,12 +53,13 @@ const db = throttle(retry(breaker(batch(
...
 ))));
 
-const minIntervalTime = parseInt(process.env.MIN_INTERVAL_TIME) || 5000;
+const minIntervalTime = parseInt(process.env.MIN_INTERVAL_TIME) || 30000;
 const maxIntervalTime = parseInt(process.env.MAX_INTERVAL_TIME) || 240000;
 const guidMinAge = parseInt(process.env.GUID_MIN_AGE) || 60000;
 const orgsToReport = process.env.ORGS_TO_REPORT ?
   JSON.parse(process.env.ORGS_TO_REPORT) : undefined;
 const lastRecordedGuid = process.env.LAST_RECORDED_GUID;
+const resultsPerPage = parseInt(process.env.PAGE_SIZE) || 100;
```
hsiliev: We already have `PAGE_SIZE` for DB paging and thus cannot change both independently.
What about prefixing with `API_`?

lib/cf/applications/src/index.js
```diff
@@ -380,7 +381,8 @@ const reportAppUsage = (cfToken, abacusToken, { failure, success }) => {
   }
 
   errors.missingToken = false;
-  let uri = '/v2/app_usage_events?order-direction=asc&results-per-page=50';
+  let uri = '/v2/app_usage_events?order-direction=asc&results-per-page=' +
```
hsiliev: We can use templates: ` /v2/app_usage_events?order-direction=asc&results-per-page=${resultsPerPage}`

lib/cf/applications/src/test/api-test.js
```diff
@@ -83,11 +83,11 @@ describe('Admin API', () => {
           applications: {
             config: {
               secured: false,
-              minIntervalTime: 5000,
```
hsiliev: Does it make sense to add `API_PAGE_SIZE` to the list of variables we test for change?

Issue #633, opened 2017-05-14
Support watch and filters in mocha


hsiliev:
The -watch argument and filters for tests makes it easier to develop & debug. 

Flags also guard against leaving exclusive tests with `.only`. We now have warning in such case, but this still does not eliminate the occasional commits that leave tests disables.  If we have filters we can simply disable the use of `.only` by setting it as error in eslint and use filters instead to achieve the same.

The two flags should be supported by mocha module.

hsiliev:
--grep and --fgrep are supported via https://github.com/cloudfoundry-incubator/cf-abacus/commit/8a7e6bca92567ccbc82ef1988e8a6b1a194a7e3b

watch cannot be supported at present without implementing duplicating/same functionality in abacus-mocha

PR #742, opened 2017-09-21
Add file option to abacus-mocha

georgi-lozev:

### DIFF
--- a/tools/mocha/src/index.js
+++ b/tools/mocha/src/index.js
@@ -53,6 +53,8 @@ const runCLI = () => {
   }
   else {
     commander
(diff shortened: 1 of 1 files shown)


hsiliev:
Can you elaborate on why we need the file option? What is the use-case behind it?

georgi-lozev:
Not exactly, we'd like to support execution of concrete files and only on all files under test directory which end with `test.js`.

The https://github.com/cloudfoundry-incubator/cf-abacus/issues/633 seems to be more flexible, but it's more kind of improvement and will require additional changes.

PR #746, opened 2017-09-25
Run files ending with test.js when --command option is used

georgi-lozev:


georgi-lozev:
Currently we're skipping itests execution, because they are run with`--command` option. This commit should fix it.

PR #762, opened 2017-10-12
Use all processors for workers on CF by default

hsiliev:


hsiliev:
This would defeat the purpose of in-memory locks to synchronize DB access.

PR #770, opened 2017-10-19
Make cluster worker fine tunable

nmaslarski:


lib/utils/cluster/src/test/test.js
```diff
@@ -257,51 +259,89 @@ describe('abacus-cluster', () => {
     expect(process.exit.args.length).to.not.equal(0);
   });
 
-  it('forks a default of 1 worker', () => {
+  context.only('calculate workers', () => {
```
hsiliev: We need to remove `.only`

PR #780, opened 2017-10-25
Implement configurable count window's granularity and size  in perf module

nmaslarski:
Signed-off-by: Martin Aleksandrov <martin.d.aleksandrov@gmail.com>


lib/utils/perf/src/index.js
```diff
@@ -19,6 +19,37 @@ const find = _.find;
 // Setup debug log
 const debug = require('abacus-debug')('abacus-perf');
 
+const countSample = process.env.PERF_COUNT_SAMPLE ? 
+  JSON.parse(process.env.PERF_COUNT_SAMPLE) : 
+  { granularity: 'seconds', size: 10 };
+
+const getWindow = (sample) => {
+  const seconds = 1000;
```
hsiliev: Do we need to calculate these constants on every invocation of getWindow?

lib/utils/perf/src/test/healty-test.js
```diff
@@ -0,0 +1,131 @@
+'use strict';
```
hsiliev: Shouldn't this file be named healthy-test.js and not healty?

lib/utils/perf/src/index.js
nmaslarski: We call it only once, on initialization

PR #791, opened 2017-11-07
Log errors in breaker

nmaslarski:


lib/utils/breaker/src/index.js
```diff
@@ -65,6 +65,9 @@ CircuitBreakerError.prototype = Error.prototype;
 // Handle a call back to an application callback
 const callback = (name, err, val, time, timeout, reject, circuit, cb) => {
 
+  if(err)
+    edebug('Error in braker: %j with err.nobreaker: %j', err, err.nobreaker);
```
hsiliev: This should read: Error in br**e**aker
Not sure if we should include also the `name` passed as callback arg

PR #804, opened 2017-11-14
Use system token for auth/authz between collector and provisioning

PlamenDoychev:


lib/metering/collector/src/index.js
```diff
@@ -363,14 +362,12 @@ const collector = () => {
...
-  // Create the app
-  const app = collector();
+const startApp = (systemToken) => {
+  const app = collector(systemToken);
   app.listen();
 
-  // Replay any failed events on master
   if (!cluster.isWorker() || cluster.isDisabled()) {
-    const replayDelay = err ? 20000 : 0;
+    const replayDelay = 1000;
```
hsiliev: Do we really need the replay delay and `setTimeout`? The code will lock much cleaner without them

PR #819, opened 2017-12-01
Enable blue-green deployment with cfpush

HristoDobchev:
Implements changes that make blue-green deployment possible. 
- cfpush now has a new command line property '-z, --prepare-zdm [boolean]' - if set to 'true' the app will be deployed without downtime by renaming the original app to <appname>-old
- the option can be provided in the application manifest - "zdm": true  
- there is a clean-up script for concourse that will remove the old application versions

### DIFF
--- a/tools/cfpush/src/index.js
+++ b/tools/cfpush/src/index.js
@@ -103,7 +103,7 @@ const prepareZdm = (props, cb) => {
     const appName = `${props.prefix}${props.name}`;
     const command = `cf app ${appName}`;
     executeCommand(command, (code) => {
-      if (code === 1)
+      if (code !== 0)
         return cb();
       return deleteOld(props, (error) => {
         if (error)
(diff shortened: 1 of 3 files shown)


hsiliev:
@HristoDobchev Can we unify the manifest & command line option and have the same name for both entries?

tools/cfpush/src/index.js
```diff
@@ -82,6 +79,43 @@ const push = (props, cb) => {
...
+  const command = `cf delete -f ${appName}-old`;
+  executeCommand(command, cb);
+};
+
+const prepareZdm = (props, cb) => {
+  if (props.prepareZdm) {
+    const appName = `${props.prefix}${props.name}`;
+    const command = `cf app ${appName}`;
+    executeCommand(command, (code) => {
+      if (code === 1)
```
hsiliev: Shouldn't this be `!== 0`

Issue #832, opened 2017-12-19
batch.groupBy fails all groups in case of error in one


georgi-lozev:
With the current implementation of `groupBy` in [abacus-batch](https://github.com/cloudfoundry-incubator/cf-abacus/blob/5181a7a6f1092fc2b0c7865c6b717764987d8014/lib/utils/batch/src/index.js#L161) event if there is is an error in a call in one of the groups, all  groups response with that same error. 

Imagine the following situation.
Two clients post their usage to the collector in parallel.  The first one with valid usage and the second one with out of slack window. The requests are received in batch in the accumulator. It starts to groupReduce them and the second one which is not in the slack throws. This will result in reporting the same error to both clients, despite the fact the usage doc of the first client has been processed successfully. This could also happen for all kind of throw errors that are not handled in the dataflow reducers, not only for out of slack.

PR #833, opened 2017-12-19
Do not fail all groups just the one that has actually an error

georgi-lozev:
For more details #832

Signed-off-by: Hristo Dobtchev <hristo.dobtchev@sap.com>


georgi-lozev:
This PR should minimize the impact of #832 

This is not the best option, because we still could have situation that there is only a problem in one of the calls in a certain group, but still the whole group will fail. However, achieving this will require changes in other modules(like transform.map) and we're not sure what will be the impact.

lib/utils/batch/src/index.js
```diff
@@ -157,8 +160,11 @@ const groupfn = (fn, group) => {
         transform.map(
           groups,
           (gmaps, i, groups, mcb) => {
-            // Return a zip of the call requests and corresponding results
-            fcb(map(gmaps, (gmap) => gmap.call), (err, gres) => err ? mcb(err) : mcb(undefined, zip(gmaps, gres)));
+            // Return a zip of the call requests and corresponding results.
+            // In case of error in any, mark all requests within the current group as failed.
+            fcb(map(gmaps, (gmap) => gmap.call), (err, gres) => {
+              return err ? mcb(undefined, zip(gmaps, map(gmaps, (gmap) => [err]))): mcb(undefined, zip(gmaps, gres));
```
hsiliev: Might be better to have a function `markAllFailed` instead of comment

lib/utils/batch/src/index.js
```diff
@@ -121,7 +121,10 @@ const batchifyfn = (fn, delay, max, size) => {
 };
 
 // Return a function that organizes a batch of calls in groups, applies each
-// group of calls and assembles the results
+// group of calls and assembles the results.
+// Warning: In case there is an error in one of the groups the whole group fails. This behaviur
+// is due the fact that with the current implementation of transform.map we could not easily distinguish where
+// the error has happened. This logic needs to be revised in future.
```
hsiliev: We need to fix `behaviur` to `behaviour` and we can replace "This logic needs to be revised in future." with github issue reference

PR #856, opened 2018-01-15
Remove cf pack

### LINKED ISSUE #821: Consider removing cfpack
We use cfpack for:
* packing internal packages
* zipping all dependencies
* fixing shrinkwrap file

Node.js buildpack for Cloud Foundry supports vendoring dependencies by simply installing them. Combined with npm 5 lock file this should give us pretty much the same functionality as cfpack.

We can therefore remove it and replace it with `npm install`  instead

denicaM:
Remove cfpack module and use npm install instead
    Fixes #821 
    [delivers #153326502]


bin/boshrelease
```diff
@@ -9,7 +9,7 @@ if [ -n "$1" ]; then
 fi
 
 for MODULE in $(node_modules/abacus-etc/apprc node_modules/abacus-etc/apps.rc $CONF apps); do
-  (cd $MODULE && npm run cfpack)
+  (cd $MODULE && npm install)
```
hsiliev: We have to delete bosh related files since we no longer have bosh release

PR #857, opened 2018-01-17
Revert "Add async functions support in router (#854)"

denicaM:
This reverts commit 9f17c9b0becd37eac425fac96d292012bf0e59e3.


hsiliev:
The commit causes problems with node 6. Until we bump all modules to have minimal version of 8 we should not reintroduce this feature

PR #894, opened 2018-03-20
Configure connection timeout in webapp

astefanova:


lib/utils/webapp/src/index.js
```diff
@@ -80,11 +80,11 @@ const webapp = () => {
         oauth.basicStrategy(process.env.API, healthCheckScopes, process.env.JWTKEY, process.env.JWTALGO)
       );
     app.get('/healthcheck', healthCheckMiddlware);
-
+
```
hsiliev: We don't need the white-space to change I think

lib/aggregation/reporting/src/index.js
```diff
@@ -1276,7 +1276,10 @@ const runCLI = () => {
     systemToken.start();
   }
 
-  reporting().listen();
+  const server = reporting().listen();
+  
+  if (process.env.CONNECTION_TIMEOUT)
+    server.setTimeout(process.env.CONNECTION_TIMEOUT);
```
hsiliev: Instead of having this in only reporting we can add it to webapp module:
https://github.com/cloudfoundry-incubator/cf-abacus/blob/9c26ee6d9433ddb971adf9926a1b9ad95bed6aeb/lib/utils/webapp/src/index.js#L86

This will also allow us to test it with a simple webserver that has the connection timeout set to under a second and expect an error.

PR #896, opened 2018-03-23
Use abacus default healthcheck middleware

PatilSukhil:
This change enabled dashboard to use default healthcheck middleware.

### DIFF
--- a/lib/cf/dashboard/src/test/authmiddleware-test.js
+++ b/lib/cf/dashboard/src/test/authmiddleware-test.js
@@ -125,6 +125,14 @@ describe('lib', () => {
         sinon.assert.calledWith(passportAuthStub,req,res,next);
         done();
       });
+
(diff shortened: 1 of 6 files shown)


lib/cf/dashboard/src/test/authmiddleware-test.js
```diff
@@ -125,6 +125,14 @@ describe('lib', () => {
         sinon.assert.calledWith(passportAuthStub,req,res,next);
         done();
       });
+
+      it('should not authenticate /healthcheck call', (done) => {
```
hsiliev: The rest of Abacus healthchecks use basic strategy for authentication. Can we adopt the same here?

PatilSukhil: Yes, we can adopt the same strategy for dashboard also.

PR #899, opened 2018-03-27
Handle update event

vgpetkov:
- Introduce Event Mapper - single to multiple events for both app and service bridge
- Refactor event converters for both app and service bridge
- Introduce Preceding Events Reader for service bridge - reads from Carry Over
- Introduce integration tests for UPDATED event:
   * handles UPDATED event of Create->Update event stream
   * skips UPDATED event of Create->Delete->Update event stream
   * handles cases when Abacus is down


lib/cf/services/src/preceding-usages-reader.js
```diff
@@ -0,0 +1,74 @@
...
+
+    const startId = 't/' + dbclient.pad16(moment.utc().valueOf());
+    const endId = 't/' + dbclient.pad16(moment.utc().startOf('month').subtract(1, 'month').valueOf()) + 'ZZZ';
+    const descending = true;
+    const opts = { startId, endId, pageSize, descending };
+    
+    const processPageFn = (usageDocs, cb) => {
+      debug(`Processing page of ${usageDocs && usageDocs.length ? usageDocs.length : 0} documents`);
+      for(let usage of usageDocs) {
+        if(foundPrecedingUsage)
```
hsiliev: Do we need another beak?

PR #921, opened 2018-05-10
Async abacus

svgradev:


hsiliev:
We need to push a fix to allow aggregate to result in 0 before merging this PR

PR #940, opened 2018-06-07
fix script for provisioning integration test

denicaM:


denicaM:
This test need CF environment.

PR #967, opened 2018-07-04
Extract aggregator models

vgpetkov:


lib/aggregation/aggregator/src/lib/models.js
```diff
@@ -0,0 +1,283 @@
...
+  });
+};
+const newSpace = function(id) {
+  return new Space(id);
+};
+
+Space.prototype.consumer = function(id, doctime) {
+  // Construct or retrieve the consumer object
+  if (!this.consumers)
+    this.consumers = [];
```
hsiliev: Please add `edebug` here

lib/aggregation/aggregator/src/lib/models.js
```diff
@@ -0,0 +1,283 @@
...
+Space.Resource = function(id) {
+  extend(this, {
+    resource_id: id,
+    plans: []
+  });
+};
+
+Space.prototype.resource = function(id) {
+  if (!this.resources)
+    this.resources = [];
```
hsiliev: Please add `edebug` here

PR #974, opened 2018-07-10
Add xdebug log to see current inflight requests per resourceId

astefanova:


lib/utils/express/src/lib/middleware/inflight.js
```diff
@@ -115,6 +133,11 @@ const inflight = (maxClientInflight, maxRemainingInflight, secured, systemEnable
     };
     res.on('finish', done);
     res.on('close', done);
+
+    /* The values in the map may not be exact because operations contained in increment/decrement are not atomic.
+       They conatain get/set and additon/substraction */
+    xdebug(`Inflight requests per resourceId ${requestsPerResourceIdsMap.toString()}`);
```
hsiliev: Do we really need xdebug on every connection? Especially if the values are not expected to reflect 100% the reality. Can we have one xdebug when there is inflight limit reached instead?

astefanova: Moved ✅

PR #978, opened 2018-07-17
Change SLACK semantics to "time after month end"

hsiliev:
Last month usage can be submitted up to X time (seconds, minutes, days) after month ended


hsiliev:
As SLACK is the only mechanism to get rid of messages retried in RabbitMQ queue we need to be careful with this PR

PR #989, opened 2018-07-26
fix(abacus-rabbitmq) handle messages existing in queue before dead le…

vgpetkov:
…tter queues

### DIFF
--- a/lib/utils/rabbitmq/src/lib/consumer.js
+++ b/lib/utils/rabbitmq/src/lib/consumer.js
@@ -55,12 +55,14 @@ class Consumer {
       setupDeadLetterQueue(channel, conf.deadLetterQueues[1]),
       channel.prefetch(conf.mainQueue.prefetchLimit),
       channel.consume(conf.mainQueue.name, async(rabbitMessage) => {
-        const message = this.messageParser.toJSON(rabbitMessage);
+        let message = this.messageParser.toJSON(rabbitMessage);
         try {
           await messageHandler.handle(message);
           await channel.ack(rabbitMessage);
         } catch (err) {
           edebug('Failed to process message, returning to queue, error %o', err);
+          if(!message.metadata)
+            message = { usageDoc: message, metadata: { retryCount: 0 } };
           message.metadata.retryCount = message.metadata.retryCount + 1;
           publishToExchange(channel, this.messageParser, message, this.config.deadLetterQueues);
           channel.ack(rabbitMessage);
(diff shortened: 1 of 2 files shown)


lib/utils/rabbitmq/src/lib/consumer.js
```diff
@@ -55,12 +55,15 @@ class Consumer {
...
       channel.prefetch(conf.mainQueue.prefetchLimit),
       channel.consume(conf.mainQueue.name, async(rabbitMessage) => {
-        const message = this.messageParser.toJSON(rabbitMessage);
+        let message = this.messageParser.toJSON(rabbitMessage);
         try {
           await messageHandler.handle(message);
           await channel.ack(rabbitMessage);
         } catch (err) {
           edebug('Failed to process message, returning to queue, error %o', err);
+          // TODO: remove this when there are no message in queue without metadata
```
martin-d-aleksandrov: You could write a backlog item in jira instead of TODO

lib/utils/rabbitmq/src/test/consumer-test.js
```diff
@@ -210,6 +210,29 @@ describe('Consumer', () => {
             assert.notCalled(amqpChannel.nack);
           });
         });
+
+        context('when message has no metadata', () => {
+          beforeEach(async() => {
+            fakeHandler.handle.returns(Promise.reject(new Error()));
+            fakeParser.toJSON.returns(JSON.parse('{ "someProp": "some val" }'));
```
martin-d-aleksandrov: do you need to parse string to json in order to return it? can't you just return an object - { someProp: 'some val'}

lib/utils/rabbitmq/src/lib/consumer.js
vgpetkov: https://jtrack.wdf.sap.corp/browse/CFA-8315

lib/utils/rabbitmq/src/test/consumer-test.js
```diff
@@ -210,6 +210,29 @@ describe('Consumer', () => {
             assert.notCalled(amqpChannel.nack);
           });
         });
+
+        context('when message has no metadata', () => {
+          beforeEach(async() => {
+            fakeHandler.handle.returns(Promise.reject(new Error()));
+            fakeParser.toJSON.returns({ someProp: 'some val' });
+            fakeParser.toRabbitMessage.returns(
+              new Buffer(JSON.stringify({ metadata: { retryCount: 1 }, usageDoc: { someProp: 'some val' } })));
```
martin-d-aleksandrov: You could extract the common constants

PR #997, opened 2018-08-03
fix linux-container measure units

momchil-sap:


momchil-sap:
In the `meter` functions of both plans we do `.div(1073741824)`, so clearly the measures are in `BYTE` unit and the metrics are the ones in `GIGABYTE` unit.

PR #1008, opened 2018-08-16
feat(buffer) extract configuration

vgpetkov:


lib/utils/config/src/lib/buffer-config.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+const oauth = require('abacus-oauth');
+const vcapenv = require('abacus-vcapenv');
+
+module.exports.bufferConfig = (getFromEnv) => ({
+  token: (authServerUri) => {
```
martin-d-aleksandrov: Should this logic be here? It does not seem like "configuration" - this call (.cache()) actually results in some rest requests to uaa.

lib/utils/config/src/test/buffer-config-test.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+const { bufferConfig } = require('../lib/buffer-config');
+
+describe('buffer config tests', () => {
```
martin-d-aleksandrov: These tests do not validates that the returned results are actually the expected

lib/utils/config/src/test/env-reader-test.js
```diff
@@ -0,0 +1,25 @@
...
+const { envReader } = require('./../lib/env-reader'); 
+
+describe('env reader tests', () => {
+  context('when env variable is exported', () => {
+    it('should return the env variable value', () => {
+      const expectedValue = 'test-var-value';
+      process.env.TEST_VAR = expectedValue;
+
+      const value = envReader('TEST_VAR');
+      delete process.env.TEST_VAR;
```
martin-d-aleksandrov: Isn't it better this line to go in either the tests that actually depends on it, or in a "beforeEach" function

PR #1037, opened 2018-09-12
Introduce sampler plan

svgradev:


package.json
```diff
@@ -164,6 +164,7 @@
     "abacus-publish": "file:tools/publish",
     "abacus-replace-template": "file:tools/replace-template",
     "abacus-secured-itest": "file:test/secured",
+    "abacus-services-hours-scenario-test": "file:test/scenario/metering-plans/standard-services-hours",
```
momchil-sap: Maybe `metering-plans` could be a module instead and some time in the future it would contain tests for all metering plans. In such a way, a lot of the boilerplate code can be reused.

Otherwise, once we decide to implement a new plan for applications, it would need to have its own scenario test module.

test/scenario/metering-plans/standard-services-hours/.gitignore
```diff
@@ -0,0 +1,3 @@
+node_modules/**
```
momchil-sap: I find the following .gitignore content better.

```
/node_modules
/.nyc_output
/coverage
/yarn.lock
```

Also notice that I have added the `yarn.lock` entry. I think we decided not to submit those (in sub-modules), and have them covered by the main `yarn.lock`.

test/scenario/metering-plans/standard-services-hours/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "abacus-services-hours-scenario-test",
+  "description": "Services hours scenario tests",
+  "license": "Apache-2.0",
+  "version": "1.1.3",
+  "private": true,
+  "homepage": "https://github.com/cloudfoundry-incubator/cf-abacus/test/acceptance/standard-services-hours-test/",
```
momchil-sap: You are missing a `tree/master` section in the URL. If you try to open your URL address, it will result in `404`.

test/scenario/metering-plans/standard-services-hours/src/test/test.js
```diff
@@ -0,0 +1,171 @@
...
+  authServerURL: process.env.AUTH_SERVER_URL || 'http://localhost:9882',
+  collectorURL: process.env.COLLECTOR_URL || 'http://localhost:9080',
+  reportingURL: process.env.REPORTING_URL || 'http://localhost:9088',
+  provisioningURL: process.env.PROVISIONING_URL || 'http://localhost:9880',
+  pollInterval: process.env.POLL_INTERVAL || 300
+};
+
+describe('standard services hours scenario test', () => {
+  const timestamp = moment.now();
+  const hourInMillis = 1 * 60 * 60000;
```
momchil-sap: If you are doing it with a formula, expand it all the way: `60 * 60 * 1000`. The leading `1` is optional.

test/scenario/metering-plans/standard-services-hours/src/test/test.js
```diff
@@ -0,0 +1,171 @@
...
+      measured_usage: [{
+        measure: 'duration',
+        quantity: hourInMillis * hours
+      }]
+    };
+
+    return usageDoc;
+  };
+
+  const sleep = (duration) => {
```
momchil-sap: The `sleep` and `eventually` methods are already used in the `dedup` test, unless I am mistaken. Consider a way extract these somewhere (maybe in our mocha wrapper) and make them reusable.

test/scenario/metering-plans/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "abacus-services-hours-scenario-test",
+  "description": "Services hours scenario tests",
+  "license": "Apache-2.0",
+  "version": "1.1.3",
+  "private": true,
+  "homepage": "https://github.com/cloudfoundry-incubator/cf-abacus/test/scenario/metering-plans",
```
momchil-sap: url is still invalid

test/scenario/metering-plans/package.json
```diff
@@ -0,0 +1,56 @@
...
+  ],
+  "files": [
+    ".npmrc",
+    "src/"
+  ],
+  "scripts": {
+    "acceptance": "yarn run scenario",
+    "scenario": "abacus-eslint && abacus-mocha --command",
+    "test": "abacus-eslint",
+    "generate-coverage": "true",
```
momchil-sap: Is the existence of the `generate-coverage` script a requirement?

test/scenario/metering-plans/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "pub": "publish"
+  },
+  "devDependencies": {
+    "abacus-oauth": "file:../../../lib/utils/oauth",
+    "abacus-eslint": "file:../../../tools/eslint",
+    "abacus-mocha": "file:../../../tools/mocha",
+    "abacus-moment": "file:../../../lib/utils/moment",
+    "abacus-publish": "file:../../../tools/publish",
+    "abacus-request": "file:../../../lib/utils/request",
+    "nyc": "^11.4.1",
```
momchil-sap: If we are not generating coverage for this module, do we need `nyc`?

test/scenario/metering-plans/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "uuid": "^3.2.1"
+  },
+  "nyc": {
```
momchil-sap: Do we need `nyc`?

test/scenario/metering-plans/src/test/test.js
```diff
@@ -0,0 +1,156 @@
...
+  let systemToken;
+
+  const authHeader = (token) => token ? { authorization: token() } : {};
+
+  const buildUsageDoc = (orgID, hours) => {
+    const usageDoc = {
+      start: timestamp,
+      end: timestamp,
+      organization_id: orgID,
+      dedup_id: uuid.v4(),
```
momchil-sap: Didn't we discuss to remove `dedup_id` from the test so that we don't overlap features? Is there really a need for `dedup_id`?

test/scenario/metering-plans/src/test/test.js
```diff
@@ -0,0 +1,156 @@
...
+const moment = require('abacus-moment');
+const oauth = require('abacus-oauth');
+const request = require('abacus-request');
+const util = require('util');
+const uuid = require('uuid');
+
+const doGet = util.promisify(request.get);
+const doPost = util.promisify(request.post);
+
+const env = {
```
momchil-sap: I would name this variable `cfg`, as some of the values are defaults. Furthermore, someone might get confused further down the code that we are reading environment variables there.

test/scenario/metering-plans/src/test/test.js
```diff
@@ -0,0 +1,156 @@
...
+    setEventuallyPollingInterval(env.pollInterval);
+
+    await Promise.all([
+      createMeteringMapping(resourceId, planName, meteringPlanId, systemToken),
+      createRatingMapping(resourceId, planName, meteringPlanId, systemToken),
+      createPricingMapping(resourceId, planName, meteringPlanId, systemToken)
+    ]);
+  });
+
+  context('submit usages', () => {
```
momchil-sap: Is there any reason to have this `context`? Maybe the `it`s could be moved one level up.

test/scenario/metering-plans/src/test/test.js
```diff
@@ -0,0 +1,156 @@
...
+};
+
+describe('standard services hours scenario test', () => {
+  const timestamp = moment.now();
+  const hourInMillis = 60 * 60 * 1000;
+  const resourceId = 'sampler-test-resource-id';
+  const meteringPlanId = 'standard-services-hours';
+  const planName = 'standard';
+
+  let usageDoc;
```
momchil-sap: This variable should not be global. The only place you are using it is in the `it` below.

test/scenario/metering-plans/src/test/test.js
```diff
@@ -0,0 +1,156 @@
...
+        firstDoc: 2,
+        secondDoc: 4,
+        thirdDoc: -1
+      };
+      const currentMonth = 0;
+      const monthsReport = 4;
+      const usageHoursIndex = 0;
+      const samplerTestResourceIdIndex = 0;
+
+      usageDoc = buildUsageDoc(orgID, usageHours.firstDoc);
```
momchil-sap: Don't mutate a single variable. Either inline the method calls or use names like `firstUsage`, `secondUsage`, etc.

tools/mocha/src/lib/eventually.js
```diff
@@ -0,0 +1,28 @@
...
+const setEventuallyPollingInterval = (pollingInterval) => {
+  _defaultPollingIntervalInMillis = pollingInterval;
+};
+
+const eventually = async (func, pollingInterval = _defaultPollingIntervalInMillis) => {
+  while (true) {
+    try {
+      return await func();
+    } catch (e) {
+      console.log('Eventually failed due to: %o', e.message);
```
momchil-sap: I am not sure if this should be `console.log`. It produces ugly messages.
A good approach here would be to have also a `timeout` setting. Once the eventually has retried above the `timeout` setting, you could return the last error caught by the `catch` statement.

test/scenario/dedup-id-test/package.json
```diff
@@ -28,15 +28,12 @@
     "pub": "publish"
   },
   "devDependencies": {
-    "abacus-batch": "file:../../../lib/utils/batch",
```
momchil-sap: Good job on cleaning up dependencies!

tools/mocha/src/lib/eventually.js
svgradev: We are going to change this console.log to edebug.
We were discussing to this timeout feature but we decided to rely on mocha timeout. Of course we could implement such one.

tools/mocha/src/mocha.js
```diff
@@ -35,6 +36,9 @@ const runCLI = () => {
   global.stub = global.sinon.stub;
   global.assert = global.sinon.assert;
   global.stubModule = require('./lib/stubber');
+  global.setEventuallyPollingInterval = setEventuallyPollingInterval;
```
hsiliev: It seems a big awkward to set a global polling interval. It changes the global per-module defaults. Only the fact that we execute tests in new process allows us to have different timeouts.

Having per call config makes this more explicit:
```
const lenghtyPollingConfig = {
  pollingInterval: 10000
}

const shortPollingConfig = {
  pollingInterval: 100
}

eventually(func1, shortPollingConfig)
eventually(func2, shortPollingConfig)
eventually(func2, lengthyPollingConfig)
```

I wonder can we make this more scope-bounded, like mocha test timeouts.

svgradev: Yes, this is true. But if we want to set the configuration once and execute all functions with the same config.

tools/mocha/src/lib/eventually.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
+
+const moment = require('abacus-moment');
+const edebug = require('abacus-debug')('e-abacus-mocha-eventually');
+const { defaults } = require('underscore');
+
+let _defaultPollingIntervalInMillis = 100;
+let _defaultTimeoutInMillis = 60000;
```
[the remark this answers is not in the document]
svgradev: We are not sure about this approach because it is not clear that eventually reads both environment variables and we are not able to see eventually configuration, that we are using, from the test file.

PR #1038, opened 2018-09-13
Restructure tests

astefanova:
Restructure introduces:
* **test\integration** folder:
  * contains integration tests which may depend on local MongoDB or RabbitMQ
  * run via `yarn run integration`
* **test\scenario** folder:
  * contains end-to-end scenarios using a fully deployed system
  * run via `yarn run scenario`
* **test\performance** folder:
  * contains load tests and require a pre-deployed system
  * run via `yarn run performance`
* **test\dependency** folder:
  * contains tests against remote systems (ones that Abacus depends on) and verify that their API contracts have not changed
  * run via `yarn run dependency`


test/test-utils/helper/src/index.js
```diff
@@ -0,0 +1,26 @@
...
+
+const { values, every } = require('underscore');
+
+const allValuesAreSet = (env) => {
+  return every(values(env), (value) => {
+    return typeof value !== 'undefined';
+  });
+};
+
+const checkCorrectSetup = (env) => {
```
martin-d-aleksandrov: when do we have a case with something in env being "undefined"?

test/scenario/broker/src/test/test.js
```diff
@@ -111,29 +98,46 @@ const complexMeteringPlan = {
...
+
     app = App.deploy({
       target:{
         orgName: testEnv.org,
         spaceName: testEnv.space
       },
       app: {
         name: `${moment.utc().valueOf()}-test-app`,
-        manifest: `${__dirname}/apps/test-app/manifest.yml`
+        manifest: `${getAbacusDirPath()}/test/test-utils/apps/test-app/manifest.yml`
```
martin-d-aleksandrov: I am not sure about that "hidden" dependency between modules. I'd expect that the test would get the needed resources though its dependencies and not by traversing the directory tree.

test/scenario/broker/README.md
```diff
@@ -0,0 +1,29 @@
+abacus-broker-scenario-test
+===
+
+:information_source: In order to run the broker scenario test, export the following environment variables:
+
+```bash
+export CF_API_URI=https://api.<domain>
+export CF_USER=<prefix>abacus
```
momchil-sap: Just `<username>`. The `prefix` thing should be common sense.

test/scenario/dedup-id/src/test/test.js
```diff
@@ -6,6 +6,7 @@ const request = require('abacus-request');
 const { extend } = require('underscore');
 const util = require('util');
 const uuid = require('uuid');
+
```
momchil-sap: Why the blank line? :smile:

test/smoke/healthchecker/src/test/test.js
```diff
@@ -0,0 +1,74 @@
...
+        appsResponses.every((responseCode) => expect(responseCode, appGroup + ': ' + responseCode).to.equal(200));
+      }
+      done();
+    });
+  };
+
+
+  it('reports the health of client facing components', function(done) {
+    this.timeout(testEnv.totalTimeout + 2000);
+    const testGroups = [
```
momchil-sap: In the `healthchecker` the internal and external applications are configurable. I would expect that this is the same in the `smoke` test as well so that the smoke test can be used on landscapes with various deployment schemes.

test/test-utils/abacus-util/package.json
```diff
@@ -29,13 +32,10 @@
...
   },
   "dependencies": {
-    "commander": "2.8.1"
-  },
-  "devDependencies": {
-    "abacus-eslint": "file:../../../tools/eslint",
-    "abacus-publish": "file:../../../tools/publish",
-    "abacus-rabbitmq": "file:../../../lib/utils/rabbitmq",
-    "abacus-mocha": "file:../../../tools/mocha"
+    "abacus-cmdline": "file:../../../tools/cmdline",
```
momchil-sap: This module has no tests and all of the code it exposes is expected to be used transitively by other modules. Hence, some dependencies should be in the `dependencies` section and not the `devDependencies` one.

Furthermore, the `scripts` section above contains `itest` and `test` commands which are irrelevant.

test/test-utils/cf-util/src/index.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+const cmdline = require('abacus-cmdline');
+
+module.exports = ({ api, user, password }) => {
+  const cfUtils = cmdline.cfutils(api, user, password);
+  return {
+    App: require('./cf-app')(cfUtils),
```
momchil-sap: Again, avoid dynamic `require` statements.

test/test-utils/helper/src/index.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+const { pairs, every } = require('underscore');
+
+const insufficientSetUp = (env) => {
```
momchil-sap: Keep method names consistent. Rename this method to `insufficientSetup` (lowercase `u`) so that it is consistent with the exposed method.

test/test-utils/static-apps/src/index.js
```diff
@@ -0,0 +1,31 @@
...
+    body: usageBody
+  }, cb);
+
+  return {
+    getCredentials,
+    postUsage
+  };
+};
+
+module.exports.testAppClient = testAppClient;
```
momchil-sap: Opt for a combined export.

```js
module.exports = {
  ...
};
```

test/integration/aggregation/reporting/src/test/test.js
```diff
@@ -14,20 +12,12 @@ const yieldable = require('abacus-yieldable');
...
-  .parse(argv);
-
-// External Abacus processes start timeout
-const startTimeout = commander.startTimeout || 30000;
-
-// This test timeout
-const totalTimeout = commander.totalTimeout || 60000;
+const testEnv = {
+  db: process.env.DB_URI,
+  offset: () => process.env.ABACUS_TIME_OFFSET ? parseInt(process.env.ABACUS_TIME_OFFSET) : 0,
```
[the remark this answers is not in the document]
astefanova: `process.env.ABACUS_TIME_OFFSET` is changed dynamically in the `'4 days in the future' context` so it needs to be a function.

test/scenario/broker/README.md
astefanova: It is common sense, but I would like to be able to easily copy-paste and change the values in order to run the test locally, so I decided to leave it.

test/test-utils/abacus-util/package.json
astefanova: 1) itest is removed
2) I am leaving test in order to run eslint checks
3) all modules are already in the `dependencies` section. Is there something I am misunderstanding/

PR #1053, opened 2018-09-26
Implement sampler's worker and receiver components

denicaM:


etc/apps.rc
```diff
@@ -17,22 +19,3 @@ medium:
 large:
 
 xlarge:
-
```
hsiliev: We need that profile for update-dependencies. Otherwise it does not know application modules.
If we have another way to iterate through modules (keywords?) we can live without this profile

lib/utils/moment/src/index.js
```diff
@@ -24,4 +24,25 @@ moment.toYYYYMM = (period) => {
   return time.getUTCFullYear() * 100 + time.getUTCMonth() + 1;
 };
 
+//
+// Returns a timestamp (in milliseconds since epoch), at the specified
+// year, month, day, hour, minute, second, and milliseconds.
+//
+moment.utcTimestamp = (year, month, day, hour, minute, second, millis) => {
```
hsiliev: Why is `moment.utcTimestamp` shorter or more expressive than `moment.utc([...]).valueOf()` ?

lib/utils/moment/src/index.js
```diff
@@ -24,4 +24,25 @@ moment.toYYYYMM = (period) => {
...
+  return moment.utc([year, month, day, hour, minute, second, millis]).valueOf();
+};
+
+moment.Jan = 0;
```
hsiliev: This can be transformed to `monthNumber = (monthName) => moment().month(monthName).format("M")` ?

lib/utils/moment/src/index.js
denicaM: After discussion we decided to keep the new utcTimestamp function.

lib/utils/moment/src/index.js
denicaM: We prefer to use constants.

PR #1074, opened 2018-10-12
Introduce span-dao query for incomplete spans

denicaM:


lib/sampler/storage/src/test/span-dao-test.js
```diff
@@ -359,19 +362,23 @@ describe('span-dao', () => {
       await thirdSpanPO.updateSpanProcessedInterval(thirdInterval, true, 1).verify();
     });
 
-    it('returns only spans that are sampled earlier than specified', async () => {
+    it('returns only spans that are sampled earlier than specified and end is not set', async () => {
```
hsiliev: Might be a good idea to have the test inside context('when end is not set') to have simple sentences

lib/sampler/storage/src/test/span-dao-test.js
```diff
@@ -359,19 +362,23 @@ describe('span-dao', () => {
...
       const before = moment.utcTimestamp(2018, Aug, 13, 12, 0, 0, 1);
       const spans = await dao.findIncompleteSpans(before, 0, 3);
       expect(spans.length).to.equal(1);
       expect(spans[0].target).to.deep.equal(firstTarget);
     });
 
-    it('returns multiple spans, when applicable', async () => {
-      const before = moment.utcTimestamp(2018, Aug, 13, 14, 0, 0, 1);
+    it('returns multiple spans, when applicable (end === true || processing.last_interval.end < before)', async () => {
```
hsiliev: Can we describe this with sentence "end span is ..."? Leaving the exact condition in the text is fine, but requires double maintenance if we decide to add more relations

denicaM: "returns multiple when end is set or spans are sampled earlier than specified" seems more understandable.

PR #1075, opened 2018-10-14
Refactor process usage test

svgradev:


test/smoke/process-usage/src/test/fixtures.js
```diff
@@ -0,0 +1,228 @@
+'use strict';
+
+const timewindow = require('abacus-timewindow');
+
+const { times } = require('underscore');
+
+const { fixturesCfg } = require('./env-config');
+// The scaling factor of each time window for creating the date string
+// [Second, Minute, Hour, Day, Month]
+const _slack = () =>
```
hsiliev: I think it would be better if slack and windows functions are moved to `timewindow` module?

test/smoke/process-usage/src/test/fixtures.js
```diff
@@ -0,0 +1,228 @@
...
+    [null],
+    [null],
+    [null],
+    _initWindows(win, timewindow.dimension.day),
+    _initWindows(win, timewindow.dimension.month)
+  ];
+};
+
+// Initial expected usage report for the test organization
+const createExpectedInitialReport = (orgID, summaries, quantites) => ({
```
hsiliev: The initial expected report is static. We can move it to an external file and require it. I don't think we gain much by having it inline and building it with functions.

test/smoke/process-usage/src/test/parse-report-utils.js
```diff
@@ -0,0 +1,68 @@
...
+const heavyApiCallsIndex = 2;
+const spaceIndex = 0;
+const consumerIndex = 0;
+const monthReport = 4;
+const currentMonth = 0;
+
+
+const _getCurrentMonth = (windows) => windows[monthReport][currentMonth];
+
+const _removeTandP = (report) => {
```
hsiliev: Can be named `removeConsumerMetadata` or `removeConsumerTimeAndProcessed`

test/smoke/process-usage/src/test/fixtures.js
svgradev: I am not sure for that. Could they be used in other modules with this implementation? It seems as functionality that might be used when we need to create report that we expect to get from Abacus.

test/smoke/process-usage/src/test/fixtures.js
svgradev: We may move it in separate file but we have to move with it some utility functions that build windows fields in reporting (_buildExpectedWindows). 

If we maintain it in external file every change of measures or number of usages send to Abacus is going to result in two changes - one in the test file itself and one in that external file.

test/smoke/process-usage/src/test/fixtures.js
```diff
@@ -0,0 +1,229 @@
...
+
+const { fixturesCfg } = require('./env-config');
+
+// The scaling factor of each time window for creating the date string
+// [Second, Minute, Hour, Day, Month]
+const _slack = () =>
+  /^[0-9]+[MDhms]$/.test(fixturesCfg.slack)
+    ? {
+      scale: fixturesCfg.slack.charAt(fixturesCfg.slack.length - 1),
+      width: fixturesCfg.slack.match(/[0-9]+/)[0]
```
momchil-sap: Why do we need to perform another match? Isn't the width just the substring from `0` to `length-1`. You have used the same logic for `scale`.

test/smoke/process-usage/src/test/fixtures.js
```diff
@@ -0,0 +1,229 @@
...
+      scale: timewindow.dimension.min,
+      width: 10
+    };  
+
+const _initWindows = (win, dimension) => {
+  const windows = [win];
+
+  if(fixturesCfg.windowsSizes && fixturesCfg.windowsSizes[dimension])
+    times(fixturesCfg.windowsSizes[dimension] - 1, () => windows.push(null));
+
```
momchil-sap: Unnecessary blank line and problems with `if` formatting.

test/smoke/process-usage/src/test/fixtures.js
```diff
@@ -0,0 +1,229 @@
...
+    }
+    : {
+      scale: timewindow.dimension.min,
+      width: 10
+    };  
+
+const _initWindows = (win, dimension) => {
+  const windows = [win];
+
+  if(fixturesCfg.windowsSizes && fixturesCfg.windowsSizes[dimension])
```
momchil-sap: I would extract `fixturesCfg.windowsSizes[dimension]` into a separate temporary constant.

test/smoke/process-usage/src/test/fixtures.js
```diff
@@ -0,0 +1,229 @@
...
+    ? {
+      scale: fixturesCfg.slack.charAt(fixturesCfg.slack.length - 1),
+      width: fixturesCfg.slack.match(/[0-9]+/)[0]
+    }
+    : {
+      scale: timewindow.dimension.min,
+      width: 10
+    };  
+
+const _initWindows = (win, dimension) => {
```
momchil-sap: I personally don't like the name `initWindows`. A better name could be `createInitializedWindows`. Normally I would expect `init` functions to be methods of classes and to modify the state of the object or to take the argument that needs initialization, but not to create a pre-initialized object.

test/smoke/process-usage/src/test/fixtures.js
```diff
@@ -0,0 +1,229 @@
...
+    times(timeWindows.getWindows(dimension).length - 1, () => windows.push(null));
+  }
+
+  return windows;
+};
+
+const _buildExpectedWindows = (summary, quantity) => {
+
+  const addProperty = (key, value, obj) => {
+    if (value)
```
momchil-sap: Is this check necessary? Is `_buildExpectedWindows` ever called with a null/undefined summary or quantity?

test/smoke/process-usage/src/test/fixtures.js
```diff
@@ -0,0 +1,229 @@
...
+          windows: _buildExpectedWindows(summaries.lightApiCalls, quantites.lightApiCalls)
+        },
+        {
+          metric: 'heavy_api_calls',
+          windows: _buildExpectedWindows(summaries.heavyApiCalls, quantites.heavyApiCalls)
+        }
+      ],
+      plans: [
+        {
+          plan_id: 'basic/basic-object-storage/' + 'object-rating-plan/object-pricing-basic',
```
momchil-sap: What is this string concatenation for?

test/smoke/process-usage/src/test/fixtures.js
```diff
@@ -0,0 +1,229 @@
...
+              windows: _buildExpectedWindows(summaries.lightApiCalls, quantites.lightApiCalls)
+            },
+            {
+              metric: 'heavy_api_calls',
+              windows: _buildExpectedWindows(summaries.heavyApiCalls, quantites.heavyApiCalls)
+            }
+          ],
+          plans: [
+            {
+              plan_id: 'basic/basic-object-storage/' + 'object-rating-plan/object-pricing-basic',
```
momchil-sap: Another string concatenation.

test/smoke/process-usage/src/test/fixtures.js
```diff
@@ -0,0 +1,229 @@
...
+                  windows: _buildExpectedWindows(summaries.lightApiCalls, quantites.lightApiCalls)
+                },
+                {
+                  metric: 'heavy_api_calls',
+                  windows: _buildExpectedWindows(summaries.heavyApiCalls, quantites.heavyApiCalls)
+                }
+              ],
+              plans: [
+                {
+                  plan_id: 'basic/basic-object-storage/' + 'object-rating-plan/object-pricing-basic',
```
momchil-sap: And another concatenation.

test/smoke/process-usage/src/test/fixtures.js
svgradev: This function addProperty is not needed.

test/smoke/process-usage/src/test/fixtures.js
svgradev: After a discussion we decided to stay with this implementation. Because it causes double check: 

```
const dimension = fixturesCfg.windowsSize ?  fixturesCfg.windowsSizes[dimension] : undefined;

if(dimension) 
...
```

test/smoke/process-usage/src/test/parse-report-utils.js
```diff
@@ -0,0 +1,69 @@
...
+    't',
+    'p'
+    );
+  return report;
+};
+
+const getThousandLightApiCallsWindows = (report) => _getCurrentMonth(report.resources[objectStorageIndex]
+  .plans[objectStoragePlanIdIndex].aggregated_usage[thousandLightApiCallsIndex].windows);
+  
+const getHeavyApiCallsWindows = (report) => _getCurrentMonth(report.resources[objectStorageIndex]
```
momchil-sap: Rename to `getHeavyAPICallsWindows` (`API` instead of `Api`)

momchil-sap: Do the same for the rest of the methods.

test/smoke/process-usage/src/test/parse-report-utils.js
```diff
@@ -0,0 +1,69 @@
...
+  report.spaces[spaceIndex].consumers[consumerIndex]);  
+  
+const cleanReport = (report) => _removeConsumerMetadata(
+  clone(omit(report, 'id', 'processed', 'processed_id', 'start', 'end')));  
+  
+const getThousandLightApiCallsQuantity = (report) => {
+  try { 
+    const monthReport = getThousandLightApiCallsWindows(report);
+    return monthReport.quantity;
+  } catch (e) {
```
momchil-sap: Using exceptions as control flow is not good. Consider having a separate function `reportReady` that checks that all necessary fields are available.

test/smoke/process-usage/src/test/report-comparator.js
```diff
@@ -0,0 +1,89 @@
...
+    expect(previousWindow[key]).to.not.equal(undefined, `Missing ${key} in previous month report`);
+
+    const currentValue = currentWindow[key];
+    const previousValue = previousWindow[key];
+    
+    expect(currentValue).to.not.equal(previousValue, util.format('No change in %s=%d detected', key, previousValue));
+
+    const diff = currentValue - previousValue - updatingStep;
+
+    expect(Math.abs(diff)).to.be.below(0.01, util.format(
```
momchil-sap: You should use `to.be.closeTo` instead.

test/smoke/process-usage/src/test/test.js
```diff
@@ -3,501 +3,121 @@
...
-        diff
-      );
-      expect(Math.abs(diff)).to.be.below(0.01, message);
-    }
-  };
-  checkIfNear('summary', s, currentWindow, previousWindow);
-  checkIfNear('quantity', q, currentWindow, previousWindow);
+const retrieveReport = async (orgID) => {
+  const resp = await doGet(':url/v1/metering/organizations/:organization_id/aggregated/usage', {
+    url: testEnv.reportingUrl,
```
momchil-sap: Check if `baseUrl` is supported in our `abacus-request` wrapper.

test/smoke/process-usage/src/test/test.js
```diff
@@ -3,501 +3,121 @@
...
-          '\n%s: Report after %d processed usage docs:\n%s\n',
-          moment.utc().toDate(),
-          processed,
-          util.inspect(report, { depth: 20 })
-        );
-
-        // Post usage and wait for report
-        map(usage, (u) => post(u, () => wait(report, processed, done)));
-      });
+    this.timeout(timeout + 2000);
```
momchil-sap: This should use `timeout` directly.

test/smoke/process-usage/src/test/fixtures.js
svgradev: We have been discussing to subtract current and previous reports. So that, we are going to move to external file expected diff report.

test/smoke/process-usage/src/test/fixtures.js
svgradev: When we apply the solution with subtracting reports these utility functions will become redundant in smoke test

PR #1078, opened 2018-10-16
Store previous month data in own partition

astefanova:
Previous month data is no longer stored in current month partition(s). Instead we store the data in previous month partition.

As the previous month partition might be taken offline (or deleted) we safeguard with SLACK. Data for previous month is accepted only up to SLACK window expiration.


lib/utils/seqid/src/index.js
```diff
@@ -27,6 +27,10 @@ let id = {
   t: 0,
   c: 0
 };
+
+const buildId = (time, count = 0) =>
```
hsiliev: We should add a test for `buildId` in the sequenceId module

lib/aggregation/accumulator/src/test/previous-month-test.js
```diff
@@ -0,0 +1,216 @@
...
+  db: process.env.DB
+};
+
+describe('abacus-usage-accumulator', () => {
+  let clock;
+  let accumulator;
+  let postSpy;
+  let server;
+
+  const mockCluster = () => {
```
hsiliev: We can move the mock functions into separate util and call them from this test, as well as the main test suite `test.js`

astefanova: After a brief discussion, we have decided that the value of this change would not justify the time it would need.

lib/utils/dataflow/src/index.js
```diff
@@ -174,31 +174,11 @@ const cacheAccums = (itype, idoc, odocs, okeys, ocache, odb) => {
...
-    debug('Retrieved cached accumulated output doc id %s, rev %s, %o', cdoc.id, cdoc.rev, cdoc);
-    return dbclient.undbify(cdoc);
-  }
-
-  // Retrieve the requested output from the db using a time range query
   const t = parseInt(otime);
-  const eid = dbclient.kturi(okey, lastMonth(t));
-  const sid = dbclient.kturi(okey, t + 2 * 86400000);
+  const eid = dbclient.kturi(okey, moment.utc(t).startOf('month').valueOf());
+  const sid = dbclient.kturi(okey, moment.utc(t).endOf('month').valueOf()) + 'ZZZ';
```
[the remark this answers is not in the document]
astefanova: Used it like that to comply to convention in abacus.

lib/utils/seqid/src/test/test.js
```diff
@@ -20,57 +17,96 @@ describe('abacus-seqid', () => {
...
+      it('returns unique sequential ids', () => {
+        // Get an id, expect the id counter to be 0
+        const id1 = seqid();
+        expect(id1).to.match(/-0-0-0-0/);
+
+        // Get another for the same time, expect the id counter to increase,
+        // making the id greater than the previously obtained one
+        const id2 = seqid();
+        expect(id2).to.match(/-0-0-0-1/);
+        expect(id2 > id1).to.be.true;
```
svgradev: minor: We may change this with:

```
expect(id2).to.be.above(id1);
```

to follow convention defined in the other test files

lib/utils/seqid/src/test/test.js
```diff
@@ -20,57 +17,96 @@ describe('abacus-seqid', () => {
...
-    const fn = () => seqid.sample(id, 'Y');
-    expect(fn).to.throw();
+  describe('sample', () => {
+
+    it('samples unique sequential ids', () => {
+      const id = seqid();
+      const t = parseInt(id);
+      expect(seqid.sample(undefined)).to.equal(undefined);
+      expect(seqid.sample(id)).to.equal(id);
+      expect(seqid.sample(id, 1)).to.equal('000' + t);
```
svgradev: We may consider to extract '000' in constant (commonPrefix or something more meaningful) since it is used a lot.

lib/utils/seqid/src/index.js
```diff
@@ -22,24 +21,49 @@ const pad16 = (t) => {
   return s.slice(s.length - 16) + tt.toString().substr(n.length);
 };
 
-// Return a unique time-based id string
+const validateTimestamp = (timestamp) => {
```
svgradev: This method has strange contract: if it is not valid throws, otherwise just exit. It might be better to return true or false.

lib/utils/seqid/src/test/test.js
astefanova: `seqid` returns string. `to.be.above` requires a number or a date.

lib/aggregation/aggregator/src/test/test.js
```diff
@@ -990,29 +980,15 @@ describe('abacus-usage-aggregator', () => {
       records++;
       if (records === 1) postNov();
       if (records === 3) {
-        expect(reqs[0][1].body.resources[0].plans[0].aggregated_usage[0].windows[4][1].quantity).to.deep.equal(
-          expectedMonth
-        );
-        expect(reqs[0][1].body.resources[0].plans[0].aggregated_usage[0].windows[3][1].quantity).to.deep.equal(
-          expectedDay
-        );
+        expect(reqs[0][1].body.resources[0].plans[0].aggregated_usage[0].windows[4][1]).to.be.null;
```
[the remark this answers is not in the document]
astefanova: We can and we shall - in a separate PR, since this a big change already. The refactoring can be dealt with separately.

lib/utils/seqid/src/index.js
astefanova: Before this change `seqid` used to return the next available id, using current time, calculated within. After the change, we allow `seqid` to be used with "outiside" provided sting, which can be anything, by throwing an error here, we are making sure it will not be used in a non-planned or wrong way.
If we make this method return true/false, we would just move the throwing statement.

PR #1082, opened 2018-10-17
Simplify existing java demo

hsiliev:


demo/java-demo-client/src/main/java/org/cloudfoundry/abacus/demo/model/Usage.java
```diff
@@ -0,0 +1,128 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Arrays;
+import java.util.StringJoiner;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class Usage {
+
+  private long start;
```
momchil-sap: Add `@JsonProperty` annotation to `start` and `end` for consistency and refactor-rename protection.

demo/java-demo-client/src/main/java/org/cloudfoundry/abacus/demo/model/Space.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Optional;
+import java.util.StringJoiner;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class Space {
+
+  @JsonProperty("space_id")
+  private String spaceID;
+
+  private Resource resources[];
```
momchil-sap: Add `@JsonProperty` for consistency and refactor protection.

demo/java-demo-client/src/main/java/org/cloudfoundry/abacus/demo/DemoApplication.java
```diff
@@ -0,0 +1,87 @@
...
+    postUsage(restTemplate);
+
+    Thread.sleep(2000);
+
+    Report report = getReport(restTemplate);
+
+    int calls = getMonthlyHeavyAPICalls(report);
+    log.info("Monthly heavy API calls {}", calls);
+
+    System.exit(0);
```
momchil-sap: Is `System.exit` necessary? Doesn't the app exit on its own?

demo/java-demo-client/src/main/java/org/cloudfoundry/abacus/demo/DemoApplication.java
```diff
@@ -0,0 +1,87 @@
...
+  public void run(ApplicationArguments args) throws Exception {
+    RestTemplate restTemplate = new RestTemplate();
+
+    postUsage(restTemplate);
+
+    Thread.sleep(2000);
+
+    Report report = getReport(restTemplate);
+
+    int calls = getMonthlyHeavyAPICalls(report);
```
momchil-sap: Do we need to log the monthly heavy api calls, if we either way log the whole report in the `getReport` method?

demo/java-demo-client/src/main/java/org/cloudfoundry/abacus/demo/DemoApplication.java
```diff
@@ -0,0 +1,87 @@
...
+    SpringApplication.run(DemoApplication.class, args);
+  }
+
+  @Override
+  public void run(ApplicationArguments args) throws Exception {
+    RestTemplate restTemplate = new RestTemplate();
+
+    postUsage(restTemplate);
+
+    Thread.sleep(2000);
```
momchil-sap: Is there some way to maybe poll for a non-empty report and avoid a hardcoded sleep?

demo/java-demo-client/src/main/java/org/cloudfoundry/abacus/demo/DemoApplication.java
hsiliev: As this is Spring boot app it is kept up&running. You need to call the shutdown actuator endpoint (disabled by default). Since this is meant to be example/demo console app I decided to add System.exit to simplify testing and avoid questions like "./gradlew bootRun hangs at 75%"

demo/java-demo-client/src/main/java/org/cloudfoundry/abacus/demo/DemoApplication.java
hsiliev: This is to demonstrate how to work with the report and verify that the model is useful. Without this call we wouldn't need the stream methods, but the Report class (and nested classes) would be unusable without plumbing code.

demo/java-demo-client/src/main/java/org/cloudfoundry/abacus/demo/DemoApplication.java
hsiliev: Yep - we can add 3 additional REST calls. Not sure if we have to do that in the demo. We don't recommend this in the usual Abacus POST/GET flow.

demo/java-demo-client/src/main/java/org/cloudfoundry/abacus/demo/model/Space.java
hsiliev: Done

demo/java-demo-client/src/main/java/org/cloudfoundry/abacus/demo/model/Usage.java
hsiliev: Done

.travis.yml
```diff
@@ -85,13 +83,31 @@ matrix:
...
         - mongodb
         - rabbitmq
       script:
         - yarn run provision
         - travis_retry yarn run citest
+    - name: "Java Demo"
+      language: java
+      jdk: openjdk11
+      env:
+        - DESCRIPTION=performance-tests
```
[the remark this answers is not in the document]
hsiliev: Renamed & removed redundant variables

PR #1087, opened 2018-10-22
Sampler smoke

denicaM:


test/smoke/sampler/src/test/test.js
```diff
@@ -0,0 +1,94 @@
...
+    await samplerClient.stopSampling(eventBuilder.createStopEvent(moment.now()));
+  });
+
+  it('samples usage to Abacus successfully', async () => {
+    log('Sending start event to sampler ...');
+    response = await samplerClient.startSampling(eventBuilder.createStartEvent(moment.now()));
+    expect(response.statusCode).to.be.equal(httpStatus.CREATED);
+
+    log('Sending stop event to sampler ...');
+    response = await samplerClient.stopSampling(eventBuilder.createStopEvent(moment.now()));
```
momchil-sap: I still think that you should ensure that the start and stop moments are not the same.

test/smoke/sampler/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "abacus-sampler-smoke-test",
+  "description": "Saampler smoke test",
```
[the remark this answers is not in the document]
denicaM: Done

test/smoke/sampler/src/test/test.js
```diff
@@ -0,0 +1,143 @@
...
+  after(async() => {
+    log('Stopping sampling in case of test error ...');
+    await stopSamplingGracefully(receiverClient, eventBuilder.createStopEvent(moment.now()));
+  });
+
+  it('samples usage to Abacus successfully', async () => {
+    log('Creating services mappings ...');
+    await createMappingGracefully(receiverClient, mapping);
+
+    log('Cleaning up in case of previous test error ...');
```
momchil-sap: Shouldn't this be in the `before` or `beforeEach` section? If we ever do another `it` we won't need to copy-paste it.

denicaM: New context with `beforeEach` was pushed

PR #1097, opened 2018-11-01
Add custom header in the batch request

wanglw:


hsiliev:
Can you please point out how adding `customHeader` works? Is this handled by `request` or `express` module?

hsiliev:
I don't see customHeader in the source of `request` (nor in `express`) and that's why I asked.

In fact the docs mention only `headers' option (https://github.com/request/request#custom-http-headers), so I think the change should actually be:
```
headers: r.headers
```

PR #1109, opened 2018-11-12
Process usage test subtract reports

svgradev:


test/smoke/process-usage/src/test/report-util.js
```diff
@@ -0,0 +1,52 @@
...
+
+const _removeConsumerMetadata = (report) => {
+  if(!_reportReady(report))
+    throw new Error('Empty report');
+
+  const testResourceInstanceIndex = 0;
+  const consumerIndex = 0;
+  const spaceIndex = 0;
+
+  report.spaces[spaceIndex].consumers[consumerIndex].resources[objectStorageIndex].plans[objectStoragePlanIdIndex]
```
martin-d-aleksandrov: You could extract this `report.spaces[spaceIndex].consumers[consumerIndex].resources[objectStorageIndex].plans[objectStoragePlanIdIndex].resource_instances[testResourceInstanceIndex]` complex reference as it is used twice.

test/smoke/process-usage/src/test/report-util.js
```diff
@@ -0,0 +1,52 @@
...
+
+const getStorageUsage = (report) => {
+  if(!_reportReady(report))
+    return 0;
+
+  return _getStorageWindows(report).quantity;
+};
+
+const cleanReport = (report) => _removeConsumerMetadata(
+  clone(omit(report, 'id', 'processed', 'processed_id', 'start', 'end')));
```
martin-d-aleksandrov: Do you need the "clone" invocation? It returns a shallow copy of the passed instance, which is done by "omit", too, I guess.

svgradev: Yes, it is abacus-clone which returns deep copy.

PR #1116, opened 2018-11-19
Remove launcher and webapp modules

hsiliev:
Start & kill node processes directly


lib/utils/launcher/package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "abacus-launcher",
-  "description": "Launch multiple instances of an application",
+  "description": "Launch Abacus application",
   "license": "Apache-2.0",
   "version": "1.1.3",
   "private": true,
-  "homepage": "https://github.com/cloudfoundry-incubator/cf-abacus/tree/master/lib/utils/launcher",
+  "homepage": "https://github.com/cloudfoundry-incubator/cf-abacus/lib/utils/launcher",
```
denicaM: Homepage should not be changed.

hsiliev: Fixed

PR #1120, opened 2018-11-21
Create/Update service fails in case of missing service mapping API

astefanova:


lib/cf/broker/src/routes/update-service.js
```diff
@@ -1,170 +1,101 @@
...
+    const planId = config.generatePlanId(requestParameters.instance_id, requestParameters.instance_id);
+    const userProvidedPlan = getSinglePlan(requestParameters.serviceConfig);
+
+    try {
+      if (userProvidedPlan && userProvidedPlan.resource_provider) {
+        if (!config.getMappingApi()) {
+          edebug('Cannot create service mapping. No url defined.');
+          response.status(httpStatus.BAD_REQUEST).send({
+            description: 'Provided plan is not valid. Resource provider mapping is not supported.'
+            + 'Please remove "resourceProvider" field'
```
hsiliev: Missing space before next sentence

lib/cf/broker/src/routes/create-service.js
```diff
@@ -1,238 +1,102 @@
...
-      cb(err);
+    const planId = config.generatePlanId(requestParameters.instance_id, requestParameters.instance_id);
+    const userProvidedPlan = getSinglePlan(requestParameters.serviceConfig);
+
+    try {
+      if (userProvidedPlan && userProvidedPlan.resource_provider) {
+        if (!config.getMappingApi()) {
+          edebug('Cannot create service mapping. No url defined.');
+          response.status(httpStatus.BAD_REQUEST).send({
+            description: 'Provided plan is not valid. Resource provider mapping is not supported.'
```
hsiliev: I think `Invalid plan: resource provider mapping is not supported.` fits better here.

We can ask users to remove resourceProvider **or** contact support/Abacus ops to find out why this is missing, but we have no way to know upfront if mapping API will ever be implemented/supported on a particular landscape/installation.

lib/cf/broker/src/routes/update-service.js
hsiliev: Same as above comment - we might want to rephrase this to `Invalid plan: resource provider mapping is not supported.`

lib/cf/broker/src/test/plans/plan-builder-test.js
```diff
@@ -0,0 +1,135 @@
...
+        const plan = planBuilder.createMeteringPlan(planId);
+
+        expect(plan.plan_id).to.equal(planId);
+        expect(plan.measures).to.deep.equal([{
+          name: 'sampleName',
+          unit: 'sampleUnit'
+        }]);
+
+        expect(plan.metrics.length).to.equal(1);
+        expect(omit(plan.metrics[0], 'meter', 'accumulate', 'aggregate', 'summarize')).to.deep.equal({
```
hsiliev: Can we use `expect(plan.metrics[0]).to.deep.include.members({ ... })` here? Do we need to check for additional fields not present in members?

PR #1136, opened 2018-12-05
Collector - return 'bad request' in case account or provisioning plugin returns 'not found'

martin-d-aleksandrov:


lib/cf/broker/src/index.js
```diff
@@ -24,8 +24,9 @@ const authHeaderProvider = new BearerAuthHeaderProvider({
 });
 
 const clientsFactory = {
-  createServiceMappingClient: (url) => new ServiceMappingClient(url, authHeaderProvider, skipSslValidation),
-  createProvisioningClient: (url) => new ProvisioningClient(url, authHeaderProvider, skipSslValidation)
+  createServiceMappingClient: (url) => new ServiceMappingClient(url, { authHeaderProvider, skipSslValidation }),
+  createProvisioningClient: (url) =>
+    new ProvisioningClient(url, { authHeaderProvider, skipSslValidation }, RequestStrategy.direct())
```
hsiliev: Do we want to have retry here and use the reliable strategy?

martin-d-aleksandrov: Yes, we could. However I don't want to have this "feature" implemented within this particular change as it's purpose was other (moreover the change is already big enough).
We could enhance this as well some other things (like returning 422 instead of 400 in some cases) in the future.

PR #1141, opened 2018-12-12
feat(meter) detect future usage

vgpetkov:


lib/utils/keybuilder/src/index.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+const dbClient = require('abacus-dbclient');
+
+module.exports.createMeterDocId = (usageDoc) => {
```
hsiliev: Probably we can rename the module to `docid` and have this function as`meterId` or `createMeterId`?

denicaM: I agree.

PR #1143, opened 2018-12-18
Fix event converting when previous_instance_count is not provided

PlamenDoychev:


lib/cf/applications/src/app-event-converter.js
```diff
@@ -12,7 +12,7 @@ const isPurgeEvent = (event) => {
 
 const createDefaultMeasures = (event) => {
   return {
-    previousRunningInstances: event.entity.previous_instance_count,
+    previousRunningInstances: event.entity.previous_instance_count || event.entity.instance_count,
     previousInstanceMemory: event.entity.previous_memory_in_mb_per_instance * 1048576,
```
hsiliev: We should protect previous memory as well

PlamenDoychev: Since we haven't see such an issue before we decided to default only previous_instance_count to instance_count and leave the memory without defaulting.

hsiliev: Seems legit. It would be interesting to see if we **can** have event without change in instance or memory.

PR #1154, opened 2019-01-10
fix(reporting) use end time of aggregated doc for bounds calculation

vgpetkov:


lib/utils/timewindow/src/index.js
```diff
@@ -109,11 +109,11 @@ const timeWindowBounds = (date, dim, shift) => {
 
   // Get the from bound based on the optional shift parameter
   bound.add(shift || 0, shortcut[dim]);
-  const from = moment(bound.valueOf()).toDate();
+  const from = moment(bound.valueOf()).toDate().valueOf();
```
hsiliev: Can we use `bound.valueOf()` directly instead?

lib/utils/timewindow/src/index.js
```diff
@@ -109,11 +109,11 @@ const timeWindowBounds = (date, dim, shift) => {
...
 
   // Get the to bound based on being 1 more than from in the current dimension
   bound.add(1, shortcut[dim]);
-  const to = moment(bound.valueOf()).toDate();
+  const to = moment(bound.valueOf()).toDate().valueOf();
```
hsiliev: Can we use `bound.valueOf()` directly instead?

hsiliev: Are we sure that all uses of `timeWindowBounds` expect number and not Date?

